Issue #6, opened 2017-03-09
Social Sharing of First version


HimanshuAttri:
Social Media buttons for twitter, facebook etc can be added which would facilitate the sharing of the first version of the website being visited.

@markjgraham and @rchrd2 What do you think about it? I can start working on it right away, please let me know.

prasadwajekar:
@markjgraham and @rchrd2 I would like to contributre to this issue

GaurangTandon:
This seems to be fixed too. Now the popup has four social sharing buttons for Twitter, FB, G+, LinkedIn.

PR #8, opened 2017-03-13
Support for social sharing added

minu-aswanth:
This PR solves #6. I have added two buttons each for recent version and first version, for Facebook and Twitter sharing respectively. On clicking any button, a new tab opens up with the option to share for Facebook and tweet for Twitter.


rgpt:
I really value your work but  i think You could have used a single function for Social sharing
you could have written a single function instead of two

function share(url){
 	chrome.tabs.query({currentWindow: true, active: true}, function(tabs){
     	var taburl = tabs[0].url;
     	var ShareTab = url + taburl;
 	 	chrome.tabs.create({ url: ShareTab });
 	});
 }
                                                                           **INSTEAD OF** 
**FOR FACEBOOK**
`function recent_capture_FB_function(){
 	chrome.tabs.query({currentWindow: true, active: true}, function(tabs){
     	var taburl = tabs[0].url;
     	var fbShareTab = "https://www.facebook.com/sharer/sharer.php?u=https://web.archive.org/web/2/" + taburl;
 	 	chrome.tabs.create({ url: fbShareTab });
 	});
 }` 

**FOR TWITTER**
function recent_capture_twitter_function(){
 	chrome.tabs.query({currentWindow: true, active: true}, function(tabs){
     	var taburl = tabs[0].url;
     	var twitterShareTab = "https://twitter.com/intent/tweet?url=https://web.archive.org/web/2/" + taburl;
 	 	chrome.tabs.create({ url: twitterShareTab });
 	});
 }


**Just take the case what if in future i want those to be shared on different social media platforms also then we should not write a separate function** 
solution : is to keep a single function for sharing.

minu-aswanth:
@rgpt , good point. I agree that in this case that this method works perfectly fine but we don't know if sharing is going to be same for all the social media platforms.

vbanos:

PR #10, opened 2017-03-13
Integrated Wayback Machine's site search

gureks:
User can now directly search any site on Wayback Machine's search


gureks:
@collinx are you referring to [this](https://web-beta.archive.org/)?

collinx:
@gureks yes, I am referring to that.

gureks:
I'll work on this but let's wait for the repository maintainers' review.

gureks:
Hi @markjohngraham 
Even after your merge, this piece of code is not there in the [branch](https://github.com/internetarchive/wayback-machine-chrome/tree/v1.3.3/). 
I'm pushing another commit to this PR so that the code is updated.

rgpt:
@markjgraham @rchrd2 @gureks 
Just to make it clear for us that 
do we want our users to **search by going to a new tab altogether** and after that if **he wants to share or to see any overview which we can be seen from extension only** then he has to open the extension again and then see those information.
Are these multiple transition good from user's perspective ?.

abhidas17695:
According to me its better if the search results are displayed inside the extension window itself instead of opening a new tab. Opening a new tab to display the results defeats the purpose of integrating the extension with the site search.

PR #11, opened 2017-03-13
Added support for Alexa ranking and Whois lookup

gureks:
Added buttons for one-click view of [Alexa rank statistics](http://www.alexa.com/siteinfo/) and [WHOIS lookup](http://www.whois.com/whois/) that opens in a new tab.


rgpt:
Do you think for alexa and who is lookup you need to write two function just with the difference of a url.
Is there any other possible functionalities that can change how both the function  `alexa_statistics_function()` and  `whois_statistics_function()` behave .
Because if not then in my opinion there should be a single function for that and we should avoid redundancy.

gureks:
@markjohngraham @rchrd2  It would be great if you could review this once?

rakesh-chinta:
yes, I too think there should a single integrated function for both features i.e " Alexa rank statistics" and "WHOIS lookup", for increased efficiency of the code execution.

vbanos:

Issue #13, opened 2017-03-15
Bootstrap's JavaScript requires jQuery


rgpt:
"**Uncaught Error**: Bootstrap's JavaScript requires jQuery
    at bootstrap.min.js:6"
If you check for Errors you will notice that bootstrap.min.js needs minified jquery script

gureks:
Issue resolved with PR #14

rgpt:
JavaScript is being used to make the extension look faster so closing this issue

PR #14, opened 2017-03-15
Added jquery.min.js

gureks:
This PR solves issue #13 
The dependency of bootstrap fulfils with this. Also, no more 404 error in the console.


rgpt:
you could have used content delivery network also for jquery because  it increases parallelism, this will keep scripts up-to-date, it ensures that a user will get geographically close response
But there are also some pros of not using cdn and downloading the files separately .
As at the time when cdn fails (Which happens rarely) our chrome extension may not work
 What do you think ?

gureks:
Taking note that we are working on an extension which resides on the device, in my opinion, an offline resource would be better rather than call it everytime from the network/cache.

rgpt:
**I posted issueno#13** and was working on it .. I was taking time to figure out which one is better considering pros and cons of both the CDN and storing as a file.
CDN is good as caching, good in performance,  it is more likely that users have it cached already and Moreover it downloads in parallel
I also tend to prefer local file as in case of CDN failure our extension may get deeply affected due to it

PR #15, opened 2017-03-18
Search feature (inside extension)

rgpt:
This adds Wayback Machine's Search feature in chrome extension
User is provided with the power to fire search query for any website. 

Search Works in conjunction with [Wayback's API](https://archive.org/help/wayback_api.php) using ajax calls

 

Let me make you more comfortable with the changes that i have done using a few simple steps:

**step 1)** 

**step2)**

whenever user searches for a url then it shows a button to user depicting the latest Timestamp of archive 
and when user presses that button then it loads all the archives saved during 2017. I still have to add the feature of on Year change Load data to calendar again

**step 3)** 

This depicts the Calendar containing all those entries when url is archived

**step4)**

you can click on the highlighted date and can see the list of entries for that particular month

**step5)** 

Now as every website have different versions and they must have been archived on different dates 
user should be able to see all those archived website at that point of archival

so you can click on the entry (ctrl+click) so you can open multiple tabs so you could see those website **wayback** 

step6) Now just take the case if i supply the search bar with invalid or url which is not archived yet

**Then a user should be shown a message(404 not found) and our extension should be capable enough to save that webpage automatically ** 

I used clndr.js for calendar, [moment.js for parsing](https://momentjs.com/) [,underscore.js ](http://underscorejs.org/) to make these things work.


vbanos:
Many thanks for your PR but this feature has been implemented in a different way.

PR #16, opened 2017-03-19
Added Search Option and Social Network Sharing

SerChirag:
Will update UI in few days


vbanos:
Thank you for your PR but we this feature has been already implemented. So, I'll close this.

PR #18, opened 2017-03-21
Add Alexa integration (shows information in the popup)

davidemdot:
I have been working on the Alexa integration with the "Wayback Machine" Chrome extension. Here is a button that shows information about the hostname, like rank and a list of related sites. The result appears in the popup (index.html). Even if necessary, I could modify the script to display the information without pressing any button, getting the data every time users accesses a new site.

It is just a demo, I am thinking on how to improve the style and to add more info.

What do you think about this?

alexa1

alexa2


mekarpeles:
Why not always show the Alexa data and not include a button?

davidemdot:
Now it shows the Alexa information when the popup has been loaded ;-).

rgpt:
I really appreciate your work but i think if we are to include "search" in wayback machine's chrome extension then we would definitely allow user to search for different websites also by providing url
so in that case as you used **window.onload** will not work as on keypress of search 

another point i have in my mind is that as we are using jquery for bootstrap so it will be good if you could have made ajax calls with jquery for that

davidemdot:
Thanks for reviewing this pull request, @rgpt.

I agree with you, but this is a code that works on the current version. If the people from the Internet Archive decide to implement new features, I will update the script to adapt it to the last version.

Regards,

david

davidemdot:
I have been working to enhance this pull request: improving style, fixing bugs (links are properly working now), adding loading image, using integrated glyphicons, and cleaning code.

Here are some screenshots showing these improvements:

alexa1

alexa2

How do you see it? @markjohngraham, @rchrd2, @mekarpeles?

PR #22, opened 2017-03-26
One-click summary view of a given site (Alexa and Whois)

abhidas17695:
This version does NOT include automatic URL archival and visual feedback to the user based on the availability of the URL in the archives.

### DIFF
--- a/scripts/popup.js
+++ b/scripts/popup.js
@@ -92,3 +92,94 @@ document.getElementById('save_now').onclick = save_now_function;
 document.getElementById('recent_capture').onclick = recent_capture_function;
 document.getElementById('first_capture').onclick = first_capture_function;
 document.getElementById('search_tweet').onclick = search_tweet_function;
+
+window.onload=function(){
+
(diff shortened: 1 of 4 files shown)


scripts/popup.js
```diff
@@ -26,4 +26,96 @@ function view_all_function(){
...
+        
+        display_alexa_info(response.sent_url);
+        display_whois_info(response.sent_url);
+    });
+    
+};
+
+function display_alexa_info(url){
+    var alexa_url = 'http://data.alexa.com/data?cli=10&dat=snbamz&url=';
+
```
rgpt: This is free api to access alexa data but as our number of users can be in huge number and this api have a throttling limit . so after sometime if the number of requests increased then it will start to throw 503 status code (Service unavailable error). so it is better if we use proper authenticated query from alexa server

abhidas17695: @rgpt  thanks for the review. I agree with you on that. The same is in case of the Whois information. However for the purpose of making a working example I felt free apis suffice.

PR #25, opened 2017-03-27
social sharing support (using single on click listener)

rgpt:
This #PR includes Social sharing option with **Nice User interface** in which button shows a little bit of animation on hover so as to attract users

In this PR i deliberately **coded in such a way that if we want to add social sharing feature for suppose linked in , and any other website then we just have to make an entry in index.html** 

as for facebook , twitter , google plus function for sharing is same but there is only a difference of url so there should be a single function to handle that.


markjohngraham:
Thank you Rishi!  Do the social sharing icons call their respective services (via JS) before one selects them?  We don't like to "Bug" our users. :-)

rchrd2:
Hello. Thank you for this pull request. Before we can merge it, can you please remove the jquery you added, and change your code to not use jquery? It should be pretty simple. I will also look over the code and leave a few other comments in line.
Thank you.

index.html
```diff
@@ -33,13 +47,22 @@
...
+            <a href="https://plus.google.com/share?url=" class="btn-social btn-sm btn-google-plus"><i class="fa fa-google-plus"></i></a>
+         </td>
+    	</tr>
+  		
     </table>
     <div style="margin: 0px; padding: 0px; border: 0px solid rgb(6, 117, 211); font-size: 19px; font-style: inherit; font-variant: inherit; font-weight: inherit; font-stretch: inherit; line-height: 1; font-family: sans-serif; vertical-align: baseline; box-sizing: content-box; overflow: unset; height: 1px; position: relative; width: auto; display: flex; background-color: rgb(9, 150, 248); color: rgb(255, 255, 255); text-align: left; align-items: center; border-radius: 3px; justify-content: center; text-decoration: none;"></div>
   </body>
 </html>
-<script src="scripts/jquery.js"></script>
+<script src="scripts/jquery.min.js"></script>
```
rchrd2: Please remove jquery. The `<script>` tag from before, was accidentally left in, and this project doesn't use jquery in order to keep the size down and performance high.

scripts/popup.js
```diff
@@ -23,6 +23,13 @@ function view_all_function(){
 	document.location.href = open_url;
 }
 
+//Sharing on Social Media Social 
+function share_on_socialMedia()
+{
+	srch_url = $('#search').val();
```
rchrd2: You are missing a `var srch_url`.
Also please remove jQuery. It should be straightforward. Thank you.

rgpt: That was silly if me sure i'll correct those tweaks.
Thank you :)

index.html
rgpt: I completely agree to you on this for performance. I also prefer java script for performance 
but then we will have error in console if you check because bootstrap's javascript needs jquery.
screenshot from 2017-03-29 11 43 40
 
i reported isse before regarding same [issue#13](https://github.com/internetarchive/wayback-machine-chrome/issues/13)
 can u please guide me on that ?

rgpt: @markjohngraham @rchrd2  should there be separate functions for social sharing buttons.
how does it bugs our users if this is not the case can u please explain

gureks:
Hi @rgpt 
I used the extension after the latest merge, the search bar for social sharing looks quite different from the sharing buttons is misleading. You working on it or can I contribute to the same?

rgpt:
@gureks i think you have created a new search bar together you could have used the same search bar and used **twitter search button** along with it.
 See the problem is that when i click on **"search for twitter"** ill be redirected to different  tab all together then all the options to see any summary info about website, social sharing all gone cause im in new tab right now ,

what can be done is we'll keep a search bar and when you want to search for tweets we can provide a utility  in the same fashion as you have done but showing like "search tweets button" 

Ill add those changes and integrate to a common search. :)

gureks:
Thanks for the update. 
Looks quite better now and is not so misleading now 👍 😄

rgpt:
@gureks  Happy to help :) Just making perfectly sure to avoid any problems whatsoever that can occur . I just wish to bring more quality to this extension.

PR #28, opened 2017-03-29
Added support for social sharing and integrated search.

anishsarangi:
**User can now search for any website in the extension (integrated  search feature).This is the UI.**
ui

**By searching any website,a button will appear showing the date when it was last archived.Clicking on it ,it will redirected to that page in another tab.**
integratedsearch
redirected_google_page

**Support for social sharing with facebook,twitter and google+.With a click on a icon,it will share the latest archived page of searched url to the that social sharing page,the user would have choosed.
More social shaing options can be added .No jquery dependancy.**
sharing_page
twitter_share
google _share

### DIFF
--- a/scripts/popup.js
+++ b/scripts/popup.js
@@ -97,3 +97,14 @@ document.getElementById('save_now').onclick = save_now_function;
 document.getElementById('recent_capture').onclick = recent_capture_function;
 document.getElementById('first_capture').onclick = first_capture_function;
 document.getElementById('search_tweet').onclick = search_tweet_function;
+window.onload=automatic_archive();
+function automatic_archive(){
+    chrome.runtime.sendMessage({message:"checkurl"},function(response)
+    {
+        console.log(response.status);
+        if(response.status===false){
+            save_now_function();
+        }
+        
+    });
+}
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -425,7 +425,29 @@ chrome.runtime.onMessage.addListener(function(message,sender,sendResponse){
           
           });
   }
-  
+  else if(message.message=='checkurl'){
+
+        chrome.tabs.query({active: true, currentWindow: true}, function(tabs){
+          var tab = tabs[0];
+          var page_url = tab.url;
+          var http=new XMLHttpRequest();
+          var new_url="http://archive.org/wayback/available?url="+page_url;
+          http.open("GET",new_url,true);
+          http.send(null);
+          http.onload=function()
+          {
+              var data=JSON.parse(http.response);
+              if(typeof data.archived_snapshots.closest =="undefined")
+              {
+                  sendResponse({status:false});
+              }
+              else
+              {
+                sendResponse({status:true});
+              }
+          }
+      });
+  }
   return true; 
 });
 


rgpt:
I really appreciate your work that you have tried to learn from mistakes in other PRs and accumulate in single but i would like to point out one or two things 

->
 ` document.getElementById('facebook_btn').addEventListener("click",function()
                    {
                         fb_share(search_url);
                     });                    
                     document.getElementById('twitter_btn').addEventListener("click",function()
                     {
                         twitter_share(search_url);
                     });              
                     document.getElementById('googleplus_btn').addEventListener("click",function()
                     {
                         google_share(search_url);
                     });`
function() parameter in addEventListener can be your own predefined function so i think there is no need for doing a function call and making other function call in it what you could have done is 
 
` document.getElementById('facebook_btn').addEventListener("click",fb_share(search_url));                    
   document.getElementById('twitter_btn').addEventListener("click",twitter_share(search_url));              
   document.getElementById('googleplus_btn').addEventListener("click",google_share(search_url));`

-> there is one if condition 
error

i think there was no need for checking `if(typeof data.archived_snapshots.closest !=="undefined")`
but you could have done in simple way is 
`if(data.archived_snapshots)`

->  i think in your implementation social sharing stops to work until you make a search for it and if it is completely correct but if user is continuously poking those sharing buttons then for them it may seem that there may be some problem with extension **(that's what i felt at the first moment)** rest is all good :) 

-> sharer links of fb, G+, twitter were designed in such a way so as they work in kind a popup but in your implementation it worked in a new tab already


What is your opinion ?

scripts/popup.js
```diff
@@ -1,29 +1,95 @@
...
+            var take_url=document.getElementById("form").value;
+            var new_url="http://archive.org/wayback/available?url="+take_url;
+            http.open("GET",new_url,true);
+            http.send(null);
+            http.onload=function()
+            {
+                var data=JSON.parse(http.response);
+                if(typeof data.archived_snapshots.closest !=="undefined")
+                {
+                    var timestamp_date=data.archived_snapshots.closest.timestamp;
```
rgpt: I think there was no need for checking `if(typeof data.archived_snapshots.closest !=="undefined")`
but what you could have done in a simple way is 
`if(data.archived_snapshots)`

anishsarangi: Yes, it would work in this case but it should be `if(data.archived_snapshots.closest)`.
But I generally avoid checking it that way as it checks for truthy values. (https://www.sitepoint.com/javascript-truthy-falsy/). e.g.

`var b = 10;
if(b) console.log(“hi”)` // Prints hi

`var b = 0;
if(b) console.log(“hi”)`  // does not print

`var b = “”
if(b) console.log(“hi”)`  // does not print

To avoid this inconsistency I normally do a type-check. But yes in this case your suggestion works. But I still stick to my style of evaluation unless there’s a forced standard.

anishsarangi:
Thanks for the comments. I appreciate you taking some time out to go through my code and providing feedback. Here are my comments/opinions-

1 - You are right. It may confuse the user. I will add `“cursor: not-allowed;”` before the search so that there will be visual hints about the disabled buttons.

2 - Yes, I have implemented them in a new tab because in my experience some major websites do it that way (e.g. Reddit, Medium). But the implementation should not matter as much as the user experience as it is just a few lines of code change for us developers. Ideally the FB share dialog (https://developers.facebook.com/docs/sharing/reference/share-dialog) should be used.

scripts/popup.js
rgpt: anyways it could be done this way my point is even if you are getting data in other format like 10, 0, "" - still their type wont be undefined as they have some type and still they get executed
and yes u could just check if data inside archived_snapshot is available or not because we are sure about the type of data inside it 
`if(typeof data.archived_snapshots.closest !=="undefined")`
i know about truthy falsy but that would be needed only when you are not so sure about the type so i think this is out of context for the JSON we are getting from wayback machine

anishsarangi: I understand your point.It's just the standard I practice to be consistent.

rgpt:
In my opinion we should never prefer facebook share dialog in extensions reason being you have to use [facebook Js SDK](https://developers.facebook.com/docs/javascript) and that would be a lot of load just for a single button 
we are not even using JQuery and using facebook sdk is not a good option. yes i agree it can be a good option but only if it is a full website 

Different views that we all share among us brings out the best in us and always helps us in creating more value to our project. Thank you

anishsarangi:
I have just tried to implement the idea of sharing through facebook using facebook share dialog.But yes,it  will  be better to use Facebook SDK for JavaScript.

vbanos:

PR #33, opened 2017-03-31
Added Overview

SerChirag:
The popup of the extension can now give brief information about the archived page.
Namely number of times the page has been archived and the date range in which this has happened.

screenshot 4
screenshot 5


rgpt:
i think in a way this implementation can make extension a little fast as you will only search for a particular date and in that way we need not have to get a chunk of data from cdn about search but what if user searches on a particular date between that range he does not find any entry (use should be notified about that ) and 

secondly, if he/she is getting continuously blank results for the dates he is asking they he/she can get pissed of and stop using our extension so there should be a way by which they get to know about the available dates too 

what is your opinion ? lets discuss about this because i am also stuck with this ...... whether to follow business logic or whether ask for specific dates or maybe change design in such a way that it could get data and still be fast

SerChirag:
@rgpt 
Thanks for commenting.
I did not get the completely get the first part of your comment, but I have made certain changes in the overview function.
Please review this PR: https://github.com/internetarchive/wayback-machine-chrome/pull/35

Second part of the comment, 
I added a "fetching data" in my new version based on your suggestion. This will hopefully convey to the user that the extension is still loading data. We can add a basic javascript preloader , but I think "fetching data" does the trick.  What are your opinions ?

markjohngraham:
Please try again.

We pushed an update that should fix the "blank page" issue you are reporting.

- Mark Graham
Director, Wayback Machine, Internet Archive

SerChirag:
@markjgraham 
I did not quite get your comment. It would mean a lot if you expanded upon what you mean.
In case you were talking about the user betting frustrated on seeing a blank page, while the data loads, it has already been fixed in https://github.com/internetarchive/wayback-machine-chrome/pull/35.

Thanks again for your time.

vbanos:
Already implemented. Thank you.

PR #36, opened 2017-04-02
facebook and twitter support for extension

prasadwajekar:
check out my popup.js and index.html for facebook and twitter support for extension


vbanos:

Issue #39, opened 2017-04-03
Provide another form of visual message instead of alert when url is not archived


rgpt:
when a url is not archived and if you try to find recent capture or first capture then it puts a alert message which supposedly is not a right way to provide Notification.
for example if i try recent capture on this [doc](https://docs.google.com/document/d/1MHCUroc4pGdUTfUZ-Ro67HVg-T46Jw07aYx0T4BnfHI/edit)

i get an alert message
![4](https://cloud.githubusercontent.com/assets/11758140/24592884/e64d8bb4-183a-11e7-8427-a1151978c5f7.PNG)

PR #35, opened 2017-04-01
UI + Overview + Visual Response + Single-Event Buttons

SerChirag:
Added Cleaner Flat UI :
Since bootstrap is a large framework and also does not give developer complete control over the display, I decided to add my own CSS.
Tried to make it look as attractive as possible.
gif2
Basic Overview:
Rather than giving just a blank field, the basic overview now reads "fetching data", hopefully not irritating the user.
Once results are loaded, the fetching data disappears.
If the page has not been archived, the "To" and "From" columns also disappear.

Added Visual Response: 
If the page has not been archived, the "Save Now" button changes it color.

Single-Even-Listener Social Media Buttons:
As pointed out in my earlier PR, the social media buttons now use a single event listener.

gif1

PR #40, opened 2017-04-03
visual message if archive is not found on clicking recent or first

rgpt:
This resolves [issue#39](https://github.com/internetarchive/wayback-machine-chrome/issues/39) 
It is not only the display message that we need to change if you check the code the two scripts background.js and pop.js are taking to each other using message passing 

and recent capture , first capture is regulated in same way. There was no proper response. 
So, i am sending response based on whether url is found in archive or not.

based on that i am notifying using chrome Notifications.

even if we do not want the user to get notified by chrome notification and inside the extension itself still the code to send the response back will work the same. We can change it as required

**previous:**

After sending response and handling from the pop.js where it is supposed to be handled.
**Now:**


scripts/popup.js
```diff
@@ -4,17 +4,45 @@ function save_now_function(){
...
 
 function first_capture_function(){
 	var wb_url = "https://web.archive.org/web/0/";
 	chrome.runtime.sendMessage({message: "openurl", wayback_url: wb_url, method:'first' }, function(response) {
+	if(!response.status){
+			notify("URL not found in wayback archives!");
+		}	
 	});
 }
+function view_all_function(){
```
collinx: @rgpt you have done a great job with the visual response. There is just this small glitch which I would like to point out. You have unintendedly created a duplicate function of `view_all_function()`.

SerChirag:
Hi !
Clever of you, @rgpt to have replaced the annoying alert message with a cleaner and more subtle Notifications.
Now lets take this up a notch, shall we?
Why not disable the recent and first version button itself when we reach a page that has not been archived. Or simply just make their display property as "none".
You can get whether the page has been archived from this PR : https://github.com/internetarchive/wayback-machine-chrome/pull/35

Overall I think it would enhance user experience.
How do you feel about it ?
Any views are welcome :)

rgpt:
Thank You @SerChirag  , i think anyhow you could enable the button by **display:block** from inspected elements if you see from developers point of view. Anyways we only want our user to get know whether this webpage is archived or not. we cannot hide everything if we do not have that archive not the social sharing button and similarly this.

in this #PR i just wanted to point out that there should be a response back to extension from content script which works with dom  and we should not use **alert** there notification can be of anytype but programatically it should work properly. 

Lets see what reviewers and others have to say on this ?

rgpt:
@markjgraham @rchrd2  Kindly review this PR :)

PR #42, opened 2017-04-04
Added feature of autoarchival of webpages that are not in the wayback machine

anishsarangi:
This feature supports **automatic archiving of webpages that are not in the wayback machine.**
By clicking on the extension icon,it will automatically save the current page if it is not archived before.


anishsarangi:
@markjohngraham @rchrd2 
Kindly spare some time for reviewing my contribution.
Any feedback is most welcome.

PR #44, opened 2017-04-04
Added the feature "Autoarchival of urls that are not in the wayback machine" modified

anishsarangi:
This supports the feature of automatic archiving of webpages that are not in previously in the wayback machine.When a user is at a webpage which is not previously archived,by clicking the extension icon,it will automatically archive the page.
kindly review this PR @markjohngraham @rchrd2

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,7 +5,8 @@
     <link rel="stylesheet" type="text/css" href="css/bootstrap.min.css">
     <link rel="stylesheet" type="text/css" href="css/font-awesome.min.css">
   </head>
-  <body style="min-width: 200px;max-width: 300px;">
+  <body style="min-width:250px; width: auto;height: auto;">
+ 
   	<table width="100%" style="min-width: 120px; ">
       <tr class="spaceUnder">
         <td>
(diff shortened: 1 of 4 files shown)


abhidas17695:
Great job @anishsarangi 
One problem I noticed is that when the new tab opens in order to save the web page suppose the user clicks the extension icon on that new tab , then the extension will try to save that web page as well. This may lead to a chain of tabs being opened one after the other. You can see the screenshot 
e1732f1e6f41127e1bed72b195a43e12
 This is an extreme use case but our code should be robust enough to handle all possible use cases.
The easiest way of solving this would be to add another check to line 435.
Just add " && isValidUrl(url) " to the if condition in line 435 and this problem will be fixed.

anishsarangi:
Thank you for your review @abhidas17695 .
In my last commit,I have made changes to the code.Now the problem has been solved.

anishsarangi:
@abhidas17695 @markjohngraham 
Anymore feedback is welcome.

anishsarangi:
Now If the user is on a page that's is not previously archived and he clicks on the extension icon, it will automatically save the page in the **background** and **shows a small notification,after the page is archived**.
Previously it was saving the page by opening in a new tab,but now it will work in the background.The notification window will fadeout in few seconds.
Here is the working results.
**1.User is on a page**
page_arc
**2.He clicks on the extension icon.**
**3.He gets a notification about auto-archival.**
pageshown
**4.Checking if it is correctly archived or not in the wayback machine page**
result

@markjohngraham @rchrd2 kindly review this modification in PR.It is working fine.
Any feedback is most welcome.

anishsarangi:
**This feature is now working without user interaction.**
When the DOM content of the current page is loaded ,It will check if it is previously archived or not.
If not,then it will save the page automatically and will show a success notification.
@markjohngraham @rchrd2  please check this.

rakesh-chinta:
It looks like the auto-archiving error for a new tab is fixed. It has a few conflicting files, please take care of that, so we can merge it into the next version.

anishsarangi:
Conflicts have been solved. @rakesh-chinta

PR #45, opened 2017-04-05
Rectified Merge - Single search bar

rgpt:
After the merge there was a problem that there were multiple search bars.
This PR resolves it to a single search bar and i also have made minor user Interface changes so as to make eye-catching 

single search and minor ui changes

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,7 +5,7 @@
     <link rel="stylesheet" type="text/css" href="css/bootstrap.min.css">
     <link rel="stylesheet" type="text/css" href="css/font-awesome.min.css">
   </head>
-  <body style="min-width: 200px;max-width: 200px;">
+  <body style="min-width: 200px;max-width: 300px;">
   	<table width="100%" style="min-width: 120px; ">
       <tr class="spaceUnder">
         <td>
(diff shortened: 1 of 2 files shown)


rgpt:
@rchrd2 @mark Can you please look into this and merge accordingly.

markjohngraham:
Two things.

1) If I enter "iskme" in the top search form I am not able to search.

2) When I first click on the icon for the Extension I am shown this

pastedgraphic-1

rgpt:
**ABOUT SEARCH** 
there is problem with [PR#10](https://github.com/internetarchive/wayback-machine-chrome/pull/10) (pushed by gureks) which was merged but the code is not reflected in the repository 
do you want me to look into that  issue and should the search be in new tab ? 

**ABOUT POP_UP Rendering Problem** 
I am not getting this problem on my system.This is definitely problem with browser
can you please try all these things.
**1)** clear the cache of browser 
**2)** try reinstalling the chrome extension
**4)**  **chrome://flags/** do "reset all to default" (if it still does not work can you tell me if you have a flag called **GPU compositing on all pages** in flags and what is its value).
**5)** otherwise there may be some problem in chrome as similar issues has been reported in some chrome browser versions (try updating to latest version).
**6)** if problem still continues to persist please share if there is any problem in console  and version of chrome that you are using 
I tried extension on two three more systems it was working . Please do tell me if still there is any problem i will then look into it more deeply.

gureks:
@rgpt @markjohngraham I'm working on the search problem and will push a commit in my PR #10 within an hour or so.

PR #46, opened 2017-04-06
Added feature that provides a summary of a given site(Alexa Ranking)

anishsarangi:
I have been working on the feature of showing the summary of a given site (Alexa Rank).
Here is the result.
screenshot from 2017-04-06 09-09-00
**When the user is on a website and he clicks on the extension icon,he will get a small summary of the current webpage in the popup window**
screenshot from 2017-04-06 **09-09-07

### DIFF
--- a/index.html
+++ b/index.html
@@ -47,6 +47,13 @@
           <button type="button" class="btn btn-warning" id="first_capture" tabindex='-1'>First Version</button>
         </td>
       </tr>
+      <tr class="spaceUnder" allign="center">
+        <td id="show_nothing">
+          <h4 style="margin-top:0px;">Summary</h4> 
+          <pre id="show_alexa_data">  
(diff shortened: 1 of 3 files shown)


anishsarangi:
@markjohngraham @rchrd2 
Kindly review this PR.

rgpt:
you are accessing Url by message passing from the content script .
and while doing custom search from search bar it do not work so there is this gap which should be avoided.

anishsarangi:
This works for the current tab URL and shows summary of the current page.
If you want to search something in the custom search and also want the summary,then it can be done by **passing the custom searched url** in the `get_alexa_info()` as well as `url_getter()` function.
And one thing ,If you want to show a lot of things in the popup window,**It will be very clumsy** in my view.
Thank you.

PR #49, opened 2017-04-22
Better Twitter Search Results

SerChirag:
Objective: Improving the existing Twitter Search Results.
For many URL's, we do not get tweets for the URL itself. For example : 
before

By extracting keywords from the URL and using them as a backup when URL itself does not fetch any tweets, we can avoid the blank twitter search results 
after


SerChirag:
@markjgraham  @rchrd2 
Please review this :)

PR #55, opened 2017-05-31
Display Tweets about a URL (as a page popup).

rakesh-chinta:
This Pull request shows the first week BF implementation into the present version of the Internet Archive's Wayback Machine Extension for the Chrome Browser.
It automatically allows the user to search for the tweets about a given present page with the activation of a single button. It displays the most recent tweets about the page, in a separate popup instead of a new tab.


vbanos:
Already implemented.

PR #58, opened 2017-11-25
Fixed site map button, added space before URL sequence

abhidas17695:


css/RTstyle.css
```diff
@@ -16,6 +16,7 @@
 
 /* Modal Content */
 .modal-content {
+
```
vbanos: Why do you add empty space inside the CSS class definition?
In general there are many redundant spaces in this file which hinder readability.

css/sequences.css
```diff
@@ -10,6 +10,7 @@
 /*  display: block;*/
   clear: both;
   padding-top: 10px;
+
```
vbanos: The same as with previous comment. Please remove the redundant spaces.

scripts/sequences.js
```diff
@@ -295,6 +295,8 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse) {
 						var anc_arr = nodeArray;
 						// Data join; key function combines name and depth (= position in sequence).
 						var trail = document.getElementById("sequence");
+                        
+                        var textElem = document.createElement('span');
```
vbanos: You see how the identation is completely wrong here? Maybe it has something to do with your editor? Please fix it, all modifications in this file are totally off regarding identation.

vbanos:
@abhidas17695 what about his PR?
It has fixes for sitemap button and other code for context menu (!)

I suggest to make another PR to fix the sitemap button and close this.

abhidas17695:
The site map button has been fixed and the fix for the URLs was needed before the radial tree library was built. I will close this

PR #67, opened 2018-01-24
Fixed indentation and added radial tree file name to background.js

abhidas17695:
Here you go @vbanos . Indentation is fixed and I added radial-tree.umd.js to background.js to be injected. I forgot to do that in the last PR.


scripts/background.js
```diff
@@ -470,6 +475,72 @@ chrome.webRequest.onErrorOccurred.addListener(function(details) {
...
+chrome.contextMenus.create(contextMenuItemSave);
+chrome.contextMenus.onClicked.addListener(function(clickedData){
+    if(clickedData.menuItemId=='first'){
+        chrome.tabs.query({active: true, currentWindow: true}, function(tabs) {
+            var page_url = tabs[0].url;
+            var wayback_url ="https://web.archive.org/web/0/"
+            var pattern = /https:\/\/web\.archive\.org\/web\/(.+?)\//g;
+            var url = page_url.replace(pattern, "");
+            var open_url = wayback_url+encodeURI(url);
+            URLopener(open_url,url);
```
vbanos: There is a lot of code duplication. Please try to avoid this, lines 509 to 512 repeat 4 times!
I suggest to define a small utility function just before this point and use it 4 times.

vbanos: Why do we open the target page in the same tab (``URLopen``) in some cases while in other cases we open a new tab
 (``chrome.tabs.create``)?
Isn't it better to be consistent and always load content in the same tab?

abhidas17695: In all cases the target page is opened in a new tab. I have been inconsistent though with the usage of the function URLopener

abhidas17695:
@vbanos I made functions for using context menu items. Also the use of URLopener() is more consistent.

soumam007:
@abhidas17695 @vbanos I dont know your personal email id so I thought of commenting here. I came to know about Internetarchive through gsoc and want to contribute . My request to you is that it would have been better if there was a channel in gitter or in an irc so that we can communicate effectively.

PR #99, opened 2018-03-05
Auto-archival feature added and some indentation error were fixed

anishsarangi:
**When the DOM content of the current page is loaded ,it will check if it has previously archived or not.If not, then it will save the page automatically and will show a success notification(In the picture, the notification is in the bottom-right corner)**

screenshot 5

**We can also see this in the background-page console. When we are visiting some page that is not previously archived, we can see a message in the background console page of the extension that "not found before in archives " and our extension will auto-archive it .Then if we will revisit the website, we can see the message "found in archives". Means we have successfully auto-archived it :-)**
**See the below picture**
screen shot 2018-03-05 at 5 22 49 pm

@markjohngraham  @vbanos @abhidas17695  kindly review this PR.


abhidas17695:
Thank you for this pull request @anishsarangi  . If you feel any more stuff needs improvement please make the required pull requests.

anishsarangi:
Now auto archival feature is reflected in the changes in badge text of the extension(Without popping the notification).It first checks the url is previously archived or not .If it is ,then it will show "Y" in extension and if it is not then it will show "N" and then auto archived the url id done and "A" is shown in the extension .Text will be shown as shown in the down figure.
screen shot 2018-03-15 at 4 13 38 pm
 @abhidas17695 kindly see the changes

abhidas17695:
@anishsarangi what is the need for contentScript.js ?

anishsarangi:
When the current web-page is loaded,It will send a message "checkurl" as well as current page url and in background.js file where it is checking if it is previously archived or not .
I believe, It will automatically check if a webpage is auto-archived or not when the page is loaded and there will be no need to "click on extension " to see the result

abhidas17695:
I think the same functionality could be achieved without contentScript.js

anishsarangi:
@abhidas17695 .Thank you for your help .It is now working same without contentScript.js file.Kindly review it.

Issue #103, opened 2018-03-08
Value Error : font-style bold is not a font-style value 


Geekcoder555:
Description:

Value Error : font-style bold is not a font-style value 
file : sequences.css

Acceptance criteria:
Above file should be updated accordingly.

Time:
5-10 min

sankalpdayal5:
Please close the issue as this has already been solved

VishnuRaghavan:
Why isn't the issue closed?

PR #110, opened 2018-03-10
Updated as on issue #103

VishnuRaghavan:
Font style changed to bold using Font-weight property.


vbanos:
Thank you @VishnuRaghavan

PR #120, opened 2018-03-22
UI and JS Fixes

guywhogeek:
I made some major UI fixes by distributing page options on different tabs and some JS changes according to that.

Plans for future commits -
 
1. Add search bar to the top for user convenience.
2. Add CSS fonts.
3. Usage statistics monitoring.
4. Direct API integration.
5. Lint CSS and JS files.


vbanos:
I'm sorry but we cannot accept this now due to other development direction.
Thank you for your suggestions though.

PR #127, opened 2018-04-16
User can send the feedback about extension

anishsarangi:
Here i have tried to include a feedback button, by clicking on which an user can send a mail to Info@archive.org as a feedback .
screen shot 2018-04-17 at 1 45 16 am
screen shot 2018-04-17 at 1 49 28 am


anishsarangi:
screen shot 2018-04-17 at 3 49 19 am

User can give their feedback using this form and it will send an email to info@archive.org

PR #131, opened 2018-06-11
In WBM popup, URL and StatusCode are now showing without overlapping

anishsarangi:
@vbanos @markjohngraham @abhidas17695  There was a bug due to which, the URL which we were showing in WBM popup was overlapping with other content . The bug is fixed .

screen shot 2018-06-12 at 1 27 14 am


vbanos:
@anishsarangi since you are touching this part of code, please let me make an improvement suggestion.

What we have here is styling set by JS code. Can we:
a) create CSS classes with all these CSS attributes we need.
b) remove the JS code like `el.style.something = 'xxx'` 
c) remove the `style=""` attributes from HTML.
d) instead of setting different style attributes one by one, we could just set one class

The code will be MUCH cleaner and all the styling will go to one place.

anishsarangi:
@vbanos Thank you for your suggestion. I have created CSS classes and removed the styling which was done by JS code  .

vbanos:
The code seems much better now, thank you.
Two more comments:
1. The `-class` suffix is not necessary at all. Please remove it from all the classes you created.
When we define a class like `.no-more-404s-message-inner-class`, we already know its a class, it cannot be anything else. So, its redundant.
If a class name becomes too short without this (e.g. `.url`), please use another relevant keyword in that case.
2. There are redundant empty lines in the end of almost all new class declarations and between classes as well. I think its better to remove them.

anishsarangi:
@vbanos Thank you for you suggestions. But I made the naming like that for convenience .
However I have changed the classes names and also removed the redundant spaces .

css/client.css
```diff
@@ -0,0 +1,136 @@
...
+    width: 25px;
+    height: 25px;
+    border-radius: 3px;
+    background: transparent;
+    box-sizing: border-box;
+    padding: 0;
+    border: none;
+    cursor: pointer;
+}
+.imgae-button{
```
vbanos: You wanted to write `image` here?

vbanos: Its no big deal, I'm merging.

PR #133, opened 2018-06-18
Added links feature

abhidas17695:


vbanos:
@abhidas17695 some comments:
1. You don't provide any description about what the PR does. Please do that to help people understand what you do.
2. You change the VERSION number in 3 different files whenever you make an update, I'm sure that there is a way to update only `manifest.json` and then read the var from everywhere.
3. There are a lot of commented out lines like `//var link="www.google.com";` and also a lot of empty lines. Please clean them up. If a commented line is required as some kind of reference, that's ok. Keep it there but also please write a comment to explain what is it.
Thank you!

vbanos:
This is really old and out of date. I'm closing it.

PR #137, opened 2018-08-24
changed 503 interception

abhidas17695:
503, 504 etc when intercepted led to a small notification on the bottom right of the screen. This is because client.js cannot be injected into such pages. This commit makes the tab re-direct to dnserror.html on interception of 503 , 504 etc
@vbanos please check it out

### DIFF
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -80,47 +80,21 @@ chrome.webRequest.onCompleted.addListener(function(details) {
         isValidUrl(details.url)) {
               Globalstatuscode=details.statusCode;
               wmAvailabilityCheck(details.url, function(wayback_url, url) {
-              if(details.statusCode==504){
-                  //notify(wayback_url,'View an archived version courtesy of the Internet Archive WayBack Machine');
-                  chrome.notifications.create(
-                'wayback-notification',{   
-                type: 'basic', 
-                requireInteraction:true,
-                iconUrl: '/images/logo.gif', 
-                title: "Page not available ?", 
-                message:"View an archived version courtesy of the WayBack Machine",
-                buttons:[{
-                    title: "Click here to see archived version"
-                }]
-                },
-                function(id){
-                    myNotID=id;
-                } 
-              );
-              chrome.notifications.onButtonClicked.addListener(function(notifId, btnIdx) {
-    if (notifId === myNotID) {
-        if (btnIdx === 0) {
-                chrome.tabs.create({ url:wayback_url});
-                chrome.notifications.clear(myNotID);
-                myNotID=null;
-        }
-    }
-});
-              }else{
-                  chrome.tabs.executeScript(details.tabId, {
-              file: "scripts/client.js"
-            }, function() {
-              chrome.tabs.sendMessage(details.tabId, {
-                type: "SHOW_BANNER",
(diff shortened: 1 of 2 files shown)


scripts/background.js
```diff
@@ -75,47 +75,26 @@ chrome.webRequest.onCompleted.addListener(function(details) {
...
+                }
+                catch(error){
+                    chrome.tabs.update(details.tabId, {url: chrome.extension.getURL('dnserror.html')+"?wayback_url="+wayback_url+"?page_url="+url+"?status_code="+details.statusCode+"?"});
+                }
+                
+              
           }, function() {
             
-          });
+          });
```
vbanos: Could you please avoid having unwanted trailing spaces here?
Tip: vim has many ways to *highlight trailing spaces*, please look it up, its possible to do it in all major editors AFAIK.

scripts/background.js
```diff
@@ -75,47 +75,26 @@ chrome.webRequest.onCompleted.addListener(function(details) {
...
+                    page_url: details.url,
+                    status_code: details.statusCode
               });
             });
-              }
+                }
+                catch(error){
+                    chrome.tabs.update(details.tabId, {url: chrome.extension.getURL('dnserror.html')+"?wayback_url="+wayback_url+"?page_url="+url+"?status_code="+details.statusCode+"?"});
+                }
+
```
vbanos: Please avoid empty lines as well.

vbanos:
Do you have any URL handy to test it myself? Thanks!

abhidas17695:
@vbanos I have removed the trailing blank spaces. But I haven't been able to find a URL to test it on. When 503 , 504 are intercepted , there is no webpage to inject client.js into. Thus chrome.tabs.executeScript() fails. I tested this functionality few months back on wonderingsound.com. But now this URL throws 'ERR_CONNECTION_REFUSED'. So it is anyway re-directed to dnserror.html. I looked into the logs now to search for URLs which return 503 or 504. Most of them return a page so chrome.tabs.executeScript() runs and client.js is injected.

vbanos:
OK then, I'll make an artificial 503 page in my website to test it :)

abhidas17695:
Thanks a lot ! Sorry for the inconvenience

vbanos:
@abhidas17695 I'm sorry but it doesn't work for me. This is what I did to check:

1. I made an `experiment.php` in my website with a random text and archived it using SPN.
2. I changed the page to be like this:
```
header('HTTP/1.0 503 Error!');                                                  
exit;
```
Nothing happens when I visit that.

3. Then, I add a small text in the page
```
header('HTTP/1.0 503 Error!');                                                  
die("test")
```
The interception works as expected.

So, the redirect you were trying to do when there is no body in the page does not work.

Am I doing a bad test? How did you test this in the first place?

abhidas17695:
Could you give me the public URL of this web page ? I could test it myself
I hadn't tested it when I made the PR because I couldn't find a URL which returns 503 or 504 with no body
If you recall back in April or March we had discussed this feature. I had tested it on wonderingsound.com which would return 503 then. So now I couldn't test it.

abhidas17695:
I am closing this PR and making a new PR @vbanos

PR #138, opened 2018-08-24
Added documentation to the code

anishsarangi:
I have started documenting the code. 
Please check it @vbanos . Any improvement suggestion is welcomed.

### DIFF
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -1222,7 +1222,7 @@ function openThatContext(temp,url,methodOfShowing){
           });
         });
       }
-    }else{ 
+    }else{
       chrome.tabs.query({
         windowId: windowIdtest
       }, function(tabs) {
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -3,14 +3,17 @@
 * Copyright 2016, Internet Archive
 */
 var manifest=chrome.runtime.getManifest();
-var VERSION = manifest.version;      //Load version from Manifest.json file
-var Globalstatuscode="";            //Used to store the statuscode of the if it is a httpFailCodes
+//Load version from Manifest.json file
+var VERSION = manifest.version;
+//Used to store the statuscode of the if it is a httpFailCodes
+var Globalstatuscode="";
+//List of exluded URLs
 var excluded_urls = [
   "localhost",
   "0.0.0.0",
   "127.0.0.1"
-];                                 //List of exluded URLs
-var previous_RTurl="";             //Used to store the previously visited URL for site-map feature
+];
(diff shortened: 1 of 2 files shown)


vbanos:
Please don't change the version of the code, you just add comments after all.
The problem is that other people also modify the version with their latest commit .e.g. https://github.com/internetarchive/wayback-machine-chrome/pull/137 and there will be a conflict.

vbanos:
I think that its better to avoid obvious comments like:
```
windowIdSingle=0;             //Window ID of the singleWindow Context
```
Also, please don't add the comments in the same line as the code. Its becoming too long and difficult to read. Please add the comments above the code.

scripts/background.js
```diff
@@ -273,40 +274,33 @@ chrome.runtime.onMessage.addListener(function(message,sender,sendResponse){
...
             var last_index=received_url.indexOf('/');
             var url=received_url.slice(0,last_index);    //URL which will be using for alexa and whois
             var open_url=received_url;          //URL which will be needed for finding tweets
             if(open_url.slice(-1)=='/') open_url=received_url.substring(0,open_url.length-1); 
             chrome.storage.sync.get(['auto_update_context'],function(event1){
               if(event1.auto_update_context==undefined){
-                event1.auto_update_context=false;
+                event1.auto_update_context=false; //By default auto-update context is off
               }
               console.log("here");
```
vbanos: Do we need to keep this in the code? You used it while developing, its best to avoid this kind of generic msgs in the production code.

scripts/background.js
```diff
@@ -1150,7 +1140,7 @@ function openThatContext(temp,url,methodOfShowing){
   var whois_url="https://archive.org/services/context/whois?url=" + url;
   var tweet_url="https://archive.org/services/context/twitter?url="+url;
   var hoaxy_url="http://hoaxy.iuni.iu.edu/#query="+url+"&sort=mixed&type=Twitter";
-  if(methodOfShowing=='tab'){
+  if(methodOfShowing=='tab'){
```
vbanos: Watch out for unnecessary trailing spaces, vim has many ways to highlight them in order to avoid having them,
there are ways to do it in all major editors.

scripts/background.js
```diff
@@ -1318,7 +1308,7 @@ function openThatContext(temp,url,methodOfShowing){
         }
       });
     }
-  }else if(methodOfShowing=='window'){
+  }else if(methodOfShowing=='window'){ //If context is to ve showed in window
```
vbanos: there are some typos in the code in general. Please review the comments wording again.

PR #140, opened 2018-08-26
Fixed 503 interception

abhidas17695:
The try catch block was not working because the error was not being thrown. The error is stored in `chrome.runtime.lastError`. I checked for this inside the callback for `chrome.tabs.executeScript()` and if it is set to `Cannot access contents of url "chrome-error://chromewebdata/` then I displayed dnserror.html.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -2,7 +2,7 @@
   "manifest_version": 2,
   "name": "Wayback Machine",
   "description": "Reduce annoying 404 pages by automatically checking for an archived copy in the Wayback Machine.",
-  "version": "3.0.0",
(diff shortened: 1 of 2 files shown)


vbanos:
Previous relevant PR: https://github.com/internetarchive/wayback-machine-chrome/pull/137

abhidas17695:
@vbanos the latest commit should fix the problem with interception of 404s. The last commit was checking for `chrome.runtime.lastError.message`. But on pages in which client.js can be injected , `chrome.runtime.lastError` is `null`. Hence the bug. I have tested the extension this time.

PR #142, opened 2018-08-28
Added Feature to show Web News->Tv news

anishsarangi:
This is a context feature which shows Web News->Tv news. 
@vbanos Please check it . It can be tested on the URL like https://www.huffingtonpost.com/entry/alex-jones-infowars-app-apple-google_us_5b694ec3e4b0de86f4a4bc1d and user can see a button on the Pop-up window of extension

### DIFF
--- a/css/recommendations.css
+++ b/css/recommendations.css
@@ -1,24 +1,36 @@
-.wrapper {
+/* #RecommendationTray {
   display:grid;
-  grid-template-columns: repeat(auto-fill, minmax(250px, 1fr));
+  grid-template-columns:repeat(auto-fill, minmax(250px, 1fr));
+} */
+
+#RecommendationTray {
+  background-color: white;
+  padding-left: 15px;
+  padding-right: 15px;
+  margin-left: auto;
+  margin-right: auto;
 }
 
-.wrapper > div{
+#RecommendationTray > div{
   padding:5px;
-  margin:10px;
-  border-radius: 25px;
-  border: 2px solid white;
-  align-items: center;
+  width: 300px;
+  margin-left: auto;
+  margin-right: auto;
+  margin-bottom: 5px;
+  border-radius: 5px;
+  border: 1px solid black;
+  background-color: #eee;
+  text-align: center;
 }
 
-.wrapper > div > p{
+#RecommendationTray > div > p{
   width: max-content;
   display:block;
   margin-left: auto;
   margin-right: auto;
   text-align: center;
 }
-.wrapper > div > a{
+#RecommendationTray > div > a{
   text-align: center;
 }
 a > img{
@@ -77,10 +89,11 @@ h2{
     padding-right: 15px;
 }
 .rec_results{
-  border-radius: 25px;
-  border: 2px solid #eee;
-  padding: 20px;
-  background-color: #eee
+  background-color: white;
+  padding-left: 15px;
+  padding-right: 15px;
+  margin-left: auto;
+  margin-right: auto;
 }
 p{
     font-family:Gill Sans Light;
--- a/css/recommendations.css
+++ b/css/recommendations.css
@@ -1,0 +1,96 @@
+.wrapper {
+  display:grid;
+  grid-template-columns: repeat(auto-fill, minmax(250px, 1fr));
+}
+
+.wrapper > div{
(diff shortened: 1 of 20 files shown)


css/recommendations.css
```diff
@@ -0,0 +1,109 @@
...
+  padding-left: 15px;
+  padding-right: 15px;
+  margin-left: auto;
+  margin-right: auto;
+}
+
+#RecommendationTray > div{
+  padding:5px;
+  width: 300px;
+  margin-left: auto;
```
vbanos: Do you know that you can write `margin-left` `margin-right` `margin-top` and `margin-bottom` in one line using the shorthand property? https://www.w3schools.com/css/css_margin.asp
You could do that with every css attribute that has `-left`, `-right` etc. Please do that to have shorter CSS files.

css/recommendations.css
```diff
@@ -0,0 +1,109 @@
...
+    max-height: 100%;
+}
+@font-face {
+    font-family: Bebas Neue;
+    src: url("../fonts/BebasNeue.otf");
+}
+@font-face {
+    font-family: Gill Sans Light;
+    src: url("../fonts/GillSans-Light.ttf");
+}
```
vbanos: I see that you repeat some CSS rules 4-5 times in other CSS files like:
https://github.com/internetarchive/wayback-machine-chrome/blob/master/css/similarweb.css
https://github.com/anishsarangi/wayback-machine-chrome/blob/master/css/recommendations.css
etc.
There are many css rules like that.

Isn't it better to have a `common.css` which will be loaded in ALL pages that's needed so that you stop repeating CSS rules?

settings.html
```diff
@@ -44,6 +44,16 @@ <h4>This will allow you to borrow books from archive.org</h4>
                     </label>
                 </div>
             </div>
+            <div id="show_option_news"  class="col-sm-3">
+                <div class="news" style="padding-top: 5px;">
+                    <h2 style="font-family: Gill Sans Light">Show News Recommendation</h2>
+                    <h4>This will allow you to show Web News -> TV News</h4>
```
vbanos: @anishsarangi please rebase your branch to master. Always do that before submitting a PR because while you were working, other changes that affect you may have happened.

In this particular case, please see how I changed the settings texts. I suggest that you change this title to
"Show Web News -> TV News Recommendations" and drop `h4` entirely as its redundant.

scripts/popup.js
```diff
@@ -315,6 +315,32 @@ function borrow_books(){
     });
 }
 
+function show_news(){
+    chrome.tabs.query({active: true,currentWindow:true},function(tabs){
+        url=tabs[0].url;
+        tabId=tabs[0].id;
+        chrome.storage.sync.get(['news'],function(event){
+            if(event.news==true){
+                if(url.includes("https://www.huffingtonpost"||"https://www.nytimes"||"https://www.forbes"||"https://www.washingtonpost")){
```
vbanos: Why do we use this only for these sites?
I tried it with https://edition.cnn.com/2018/09/01/politics/trump-administration-summer-hell-drama-key-moments/index.html and it worked fine.

anishsarangi: I was told by Max that to use this feature for these sites only.

vbanos: Is it possible to make a list with these sites so that its possible to add more sites easily?
Then, you could concatenate the list and produce the string you have now... or maybe find another good way to check.
This code is not very nice and its going to be much uglier when more sites are added.

scripts/recommendations.js
```diff
@@ -0,0 +1,68 @@
...
+window.onloadFuncs = [get_details];
+window.onload = function(){
+  for(var i in this.onloadFuncs){
+   this.onloadFuncs[i]();
+  }
+}
+function get_details(){
+  var article = null;
+  article = getUrlByParameter('url');
+  var apiURL = "http://maxreinisch-newsrecommender.crawl.archivelab.org/recommend/" + article
```
vbanos: We need to activate the IA endpoint first and you'll need to change your PR to use that. That's required to merge this.

anishsarangi: Thank you .Then I will make a commit including all the suggested changes after the endpoint is ready !!

PR #153, opened 2018-09-17
added wikibooks function to master branch

MaxReinisch:


booklist.html
```diff
@@ -0,0 +1,37 @@
...
+  </div>
+  <BR>
+  <BR>
+  <div class="wrapper" id="resultsTray">
+  </div>
+</body>
+</html>
+<script src="scripts/booklist.js"></script>
+<script src="scripts/jquery.js"></script>
+<script src="scripts/bootstrap.min.js"></script>
```
vbanos: Why are these scripts outside the doc, shouldn't they be inside html head?

MaxReinisch: I believe I was copying the format from a different html file, but I'll move those now.

MaxReinisch: actually moving them into the head gives a TypeError, which I believe is because the html elements must be rendered before the script gets called

vbanos: No, please leave them as they are now.
I've also checked other files like `annotation.html` and `similarweb.html` and they do the same thing as you with their JS.

vbanos:
@MaxReinisch I've loaded the extension from your PR and I the "Find cited books" button does NOT appear at all.
Have you tested the extension after doing your new PR? Something seems to be wrong.

MaxReinisch:
I tested it and it worked for me using https://en.wikipedia.org/wiki/Easter_Island.  It's possible my regex fails for your version of wikipedia.

vbanos:
OK, I had to clear my cache and it worked for me as well.

PR #157, opened 2018-09-21
Added Twitter Context

anishsarangi:
@vbanos I have made first commit on Twitter Context Screen. Changes are added in this PR.

### DIFF
--- a/twitter.html
+++ b/twitter.html
@@ -1,0 +1,50 @@
+<!doctype html>
+<html>
+    <head>
+        <meta content="text/html;charset=utf-8" http-equiv="Content-Type">
+        <meta content="utf-8" http-equiv="encoding">
+        <link rel="stylesheet" type="text/css" href="css/bootstrap.min.css">
+        <link rel="stylesheet" type="text/css" href="css/font-awesome.min.css">
+        <link rel="stylesheet" type="text/css" href="css/common.css">
+        <link rel="stylesheet" type="text/css" href="css/twitter.css">
+        <title>
+            Tweets of the Current URL
+        </title>
+    </head>
+    <body>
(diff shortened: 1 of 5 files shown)


vbanos:
@anishsarangi I've found a serious bug: When I try `http://www.vbanos.gr`, the actual URL that is sent to the back-end is `vbanos.`. When I try `https://cnn.com/`, I get `cnn.co`.

1. Why do you truncate the initial `http(s)://`? Please do not process it, let the back-end do that.
2. Why do you try to truncate the last chars? Please avoid that.

vbanos:
The appearance of the tweets is not very good.
Please check this out and try to use it:
https://developer.twitter.com/en/docs/twitter-for-websites/embedded-tweets/guides/css-for-embedded-tweets

Also, why the *whole* tweet text is a direct link? It shouldn't be.
Please see how embedded tweets appear in a website using the twitter widget, e.g. http://iskme.org/ and try to make it look the same in our case.
Thank you!

twitter.html
```diff
@@ -0,0 +1,50 @@
...
+                        <div class='container'>
+                            <img class='resize_fit_center'
+                            src='images/logo.gif' />
+                        </div>
+                    <div id="horizontal_line" class="extra_margin"></div>
+                </td>
+            </tr>
+        </table>
+        <div style="text-align: center">
+            <h2><strong>Tweets containing of Current URL</strong><br></h2>
```
vbanos: Tweets containg the current URL

anishsarangi: Thank you!!

anishsarangi:
I am closing this PR and requesting a new one
Thank you !

PR #160, opened 2018-09-23
Added bydefault all settings to off

anishsarangi:
@vbanos I have added by-default all settings to off. User has to turn on each feature to see the results.
Thank you!!


vbanos:
@anishsarangi you said **ALL** settings to but you only modified news & wikibooks...
Could you please set all of them to off?

PR #163, opened 2018-09-24
Improved UI for TVNews and DOI

MaxReinisch:
Made several UI changes, especially in the DOI CS.  The next step is to make the settings window and popup window more consistent.


vbanos:
Looking good.

PR #164, opened 2018-09-25
Small changes to DOI papers

MaxReinisch:
I hid the papers from where you select which context screens you want to see, however if it is enabled in the "Read papers cited by wikipedia", it still shows up as a context screen when you click "show me context".  I would like to change this, but that would require digging through the background.js, and I am having trouble understanding a lot of it.  Could we perhaps review that file and possibly refactor some of it?


vbanos:
Refactoring is definitely required in many places in the extension code.

PR #170, opened 2018-10-02
Removed duplicate codes

anishsarangi:
@vbanos I found a serious bug while testing the extension in Show-all-ContextScreens-in-windows mode of showing context Screen. It was not auto-updating while I turned on the "Auto-update" feature .I have fixed this Bug.
Along with it I also removed duplicate code. 
Thank you !!


vbanos:
Dear @anishsarangi as I have told you before, please make 2 PRs, one for fixing the bug and another for removing duplicate code.
As it is now, I cannot see which code is the bugfix and which is the duplicate code removal.

Thank you.

vbanos:
Also, it would be better to make individual duplicate code removal PRs for specific contexts.
E.g. Make a PR to remove duplicate code in domaintools only, another one for tagcloud etc.

vbanos:
Thank you, please try to explain what exactly you try to achieve with this PR.
What is the logic behind the duplicate code removal?
Help me understand this huge PR.

anishsarangi:
I had included in every contextScreen.js file that when the window is loaded to fire that respective .
I have removed window.onload event such that that ContextScreen.js file will execute everytime. 
I have linked those files in singleWindow.html and they will execute whenever they are called.
Please let me know if anything is missing !!
Thank you!!

scripts/annotation.js
```diff
@@ -6,77 +6,65 @@ function getUrlByParameter(name){
     var length=name.length;
     return url.slice(index+length+1,indexOfEnd);
 }
-
-function get_annotations(){
```
vbanos: I'm sorry but it makes no sense to move the code outside the function. This will create too many issues.
I see that you did this in the other JS scripts as well. The same applies there

anishsarangi:
Will make another PR for this

PR #175, opened 2018-10-03
added some news sources to list for tvnews

MaxReinisch:
This required that I slightly change the way the news hosts are identified.

### DIFF
--- a/scripts/popup.js
+++ b/scripts/popup.js
@@ -333,8 +333,7 @@ function show_news(){
     chrome.tabs.query({active: true,currentWindow:true},function(tabs){
         url=tabs[0].url;
         var to_check_url=url.replace(/^https?:\/\/(www\.)?/,'');
-        // to_check_url = to_check_url.replace(/\.html/, "");
-        var news_host=to_check_url.split(".")[0]
+        var news_host=to_check_url.split(".")[0];
         tabId=tabs[0].id;
         var set_of_sites = new Set([
           "apnews",
@@ -346,12 +345,11 @@ function show_news(){
           "snopes",
           "theverge",
           "usatoday",
-          "vox", 
+          "vox",
           "washingtonpost",
         ]);
         chrome.storage.sync.get(['news'],function(event){
             if(event.news==true){
-              console.log(set_of_sites.has(news_host));
                 if(set_of_sites.has(news_host)){
                     document.getElementById('news_recommend_tr').style.display="block";
                     document.getElementById('news_recommend_tr').onclick=function(){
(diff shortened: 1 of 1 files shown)


MaxReinisch:
Your first point: This checks any url that contains ".forbes." for example, as the way the hostname is extracted is with a str.split(".").  The previous way included "www." , but that seems unnecessary.  Other than that, I don't understand how this functionality is different than how it was previously, and how you'd prefer it.  Do you want to include the ".com" or ".org" ?

Regarding point 2: Good point, I'll make that change.

vbanos:
OK, I missed that about #1.
Then, please rename the variable name. Since you extract the `hostname`, the variable name should be `final_hostname`, it shouldn't be `final_url`.

Also, please list the sites in alphabetical order, this will make them easier to read.
This becomes more helpful as the list grows longer.

MaxReinisch:
Done

scripts/popup.js
```diff
@@ -332,14 +332,27 @@ function borrow_books(){
 function show_news(){
     chrome.tabs.query({active: true,currentWindow:true},function(tabs){
         url=tabs[0].url;
-        var to_check_url=url.replace(/^https?:\/\//,'');
-        to_check_url = to_check_url.replace(/\.html/, "");
-        var final_url=to_check_url.slice(0,to_check_url.lastIndexOf(".")+1);
+        var to_check_url=url.replace(/^https?:\/\/(www\.)?/,'');
+        // to_check_url = to_check_url.replace(/\.html/, "");
+        var news_host=to_check_url.split(".")[0]
```
vbanos: missing `;`

scripts/popup.js
```diff
@@ -332,14 +332,27 @@ function borrow_books(){
...
+          "snopes",
+          "theverge",
+          "usatoday",
+          "vox", 
+          "washingtonpost",
+        ]);
         chrome.storage.sync.get(['news'],function(event){
             if(event.news==true){
-                if(list_of_sites.indexOf(final_url)>=0){
+              console.log(set_of_sites.has(news_host));
```
vbanos: Not necessary

scripts/popup.js
```diff
@@ -332,14 +332,27 @@ function borrow_books(){
 function show_news(){
     chrome.tabs.query({active: true,currentWindow:true},function(tabs){
         url=tabs[0].url;
-        var to_check_url=url.replace(/^https?:\/\//,'');
-        to_check_url = to_check_url.replace(/\.html/, "");
-        var final_url=to_check_url.slice(0,to_check_url.lastIndexOf(".")+1);
+        var to_check_url=url.replace(/^https?:\/\/(www\.)?/,'');
+        // to_check_url = to_check_url.replace(/\.html/, "");
```
vbanos: Do we need to keep this?

PR #176, opened 2018-10-05
Add First Test. 

MaxReinisch:


vbanos:
Hi @MaxReinisch ,
This PR has not only our work on unit tests but also two previous already merged commits on papers buttons and twitter. Something is not right. We need only the unit test code.
Also, please add some instructions on installing the relevant unit test packages and running the tests as we discussed last week.
Thank you!

MaxReinisch:
Something strange is going on.  I wanted to switch to the master branch to make a PR for the other changes, but several of the changes didn't revert to my fork's master branch (such as README.md, but I think that is because it was still open).  I tried `git checkout tests` and got this:
```$ git checkout tests
error: cannot stat 'test': Permission denied
error: cannot stat 'test': Permission denied
```

MaxReinisch:
Anish helped me out, I think I can get it working again now.

PR #179, opened 2018-10-09
Use webpack to bundle JS libraries

vbanos:
Create initial webpack conf and bundle jquery and bootstrap JS into
`scripts/build.js`. The aim is to bundle all external libraries there in
the future and avoid having multiple libraries referenced in source.

Use `scripts/build.js` instead of `jquery.js` and `bootstrap.min.js` in
all HTML pages.

Remove `jquery.js` and `bootstrap.min.js` from repository. They are
external libraries, they shouldn't be inside the source.

Update `package.json` with the new libraries.

Use `npm run build` to bundle JS and create `scripts/build.js`.


vbanos:
TODO for later PRs if this is fine:
1. Include any other external library we use.
2. Bundle library CSS into a `build.css` file.

PR #184, opened 2018-10-13
Added unit test to Twitter Context page

anishsarangi:
@vbanos I have tried to add unit testing to Twitter Context.
Any comments on improving the code is most welcome.


vbanos:
Thank you @anishsarangi for your PR. It is good but I'm starting to see another pattern here.

We use the SAME function in all contexts and we have the same unit test.
Can we make a `scripts/utils.js` and have common functions there starting with `getUrlByParameter`?
Then, we won't need to repeat the same code and unit tests for every context.
We'll drop the duplicate code and we'll import `getUrlByParameter` from `scripts/utils.js`.
Could you do that? Thank you.

anishsarangi:
@vbanos Please see the latest commit. Is this the change you were talking about ?

vbanos:
Yes, your change is towards the right direction.
Comments:
1. Your unit test is about a function in `utils.js` so its filename should not be `twitter.spec.js` but `utils.spec.js`.
2. Have you seen how I implemented the same function https://github.com/internetarchive/wayback-machine-chrome/blob/master/scripts/doi.js#L2 ? Its much better than the current one you use. Please use that.
3. Where else in scripts can we delete function `getUrlByParameter` and use it from `utils.js`? Please search everything and make the necessary replacements.
4. Please use `eslint` to check your code. Note that I have included `eslint` in package.json. Please do `npm install` and then use `eslint scripts/your-script.js` when you work on something to check code style.

anishsarangi:
@vbanos Great Suggestions . I have tried to implement these in the latest commit.
Question: 
1. Do we need to keep package-lock.json file?
2.I have done npm install and/but eslint is not working. When I typed eslint scripts/twitter.js, **command not found: eslint** is printed

anishsarangi:
@vbanos yes `eslint` is installed when I tried for `npm install`. Let me retry solving this issue. 
Regarding package-lock.json , I am seeing that it is changed in this PR. I have also checked that package-lock.json should be included.
Other than this, I think it is all ok.

PR #191, opened 2018-10-22
Improve annotationURL code, maintaining the same functionality

vbanos:
Reuse `scripts/annotation.js` in `annotationURL.html`.

Drop `scripts/annotationURL.js` as duplicate to `scripts/annotation.js`.

Add unit tests for annotation.

Simplify CSS.


scripts/annotation.js
```diff
@@ -1,20 +1,25 @@
...
-  var url = getUrlByParameter('url');
-  // TODO URL processing needs to be a separate function and also needs cleanup.
-  var test_url = url.replace(/^https?:\/\//,'');
-  if(test_url.includes('iskme.org')){
-    test_url=test_url.replace(/^www\./,'');
+/**
+ * Prepare hypothes.is URL to request API.
+ */
+function hypothesis_api_url(url, type) {
+  if (!/^https?:\/\//i.test(url)) {
```
anishsarangi: Can you please share some info for these two lines of code? 
Thank you.

anishsarangi: Just got the answer from here 
https://stackoverflow.com/questions/11300906/check-if-a-string-starts-with-http-using-javascript

PR #196, opened 2018-10-22
Fixed bug regarding not seeing user name in annotation_URL context

anishsarangi:
I have fixed the bug, when I try for finding Annotations of Current URL for an URL, I am not seeing the user name which I could see in Annotations of Current Domain.
I have replaced the ID by Class for which it was not working.
Thank you!!


vbanos:
Thank you for noticing and fixing it.

PR #197, opened 2018-10-22
Overview refactor

abhidas17695:
@vbanos I refactored overview.js using jQuery and added one test. There are other tests to be added for different functions. The error I mentioned of is now fixed. The test is running without any issue.

### DIFF
--- a/scripts/overview.js
+++ b/scripts/overview.js
@@ -1,7 +1,3 @@
-function getUrlByParameter (name) {
-    const url = new URL(window.location.href);
-    return url.searchParams.get(name);
-  }
 function getTotal(captures){
     var total=0;
     var data;
@@ -81,11 +77,6 @@ function get_thumbnail(){
     }});
 }
 
-window.onloadFuncs = [get_details,first_archive_details,recent_archive_details,get_thumbnail];
-window.onload = function(){
- for(var i in this.onloadFuncs){
-  this.onloadFuncs[i]();
- }
-}
-if(typeof module !=="undefined") {module.exports = {getUrlByParameter:getUrlByParameter,getTotal:getTotal};}
+
+if(typeof module !=="undefined") {module.exports = {getTotal:getTotal};}
   
--- a/scripts/overview.js
+++ b/scripts/overview.js
@@ -1,81 +1,69 @@
-//Used to extact the current URL
-function getUrlByParameter(name){
-    var url=window.location.href;
-    var indexOfEnd=url.length;
-    var index=url.indexOf(name);
-    var length=name.length;
-    return url.slice(index+length+1,indexOfEnd);
+function getUrlByParameter (name) {
+    const url = new URL(window.location.href);
+    return url.searchParams.get(name);
+  }
+function getTotal(captures){
+    var total=0;
+    var data;
+    for(var key in captures){
+        data=captures[key];
+        total=total+data['text/html'];
+    }
+    return total;
 }
-
 function get_details(){
     var url=getUrlByParameter('url');
-    var xhr=new XMLHttpRequest();
     var new_url="https://archive.org/services/context/metadata?url="+url;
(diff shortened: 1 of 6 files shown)


test/overview.spec.js
```diff
@@ -0,0 +1,16 @@
...
+        var captures = {
+            '2001': {
+                'text/html': 1,
+            },
+            '2002': {
+                'text/html': 4
+            }
+        }
+        expect(getTotal(captures)).to.equal(5);
+        expect(5).to.equal(5);
```
anishsarangi: We don't need this line of code. This can be deleted. Thank you!!

abhidas17695: Thanks for pointing that out. Its fixed.

scripts/overview.js
```diff
@@ -1,116 +1,82 @@
...
-    var indexOfEnd=url.length;
-    var index=url.indexOf(name);
-    var length=name.length;
-    return url.slice(index+length+1,indexOfEnd);
+function getTotal(captures){
+    var total=0;
+    var data;
+    for(var key in captures){
+        data=captures[key];
+        total=total+data['text/html'];
```
vbanos: This could be done in 1 line and without having the `data` var at all.
```
total += captures[key]['text/html']
```

scripts/overview.js
```diff
@@ -1,116 +1,82 @@
...
-    xhr.onload=function(){
-        if(this.response.size!=233){
-            var img = document.createElement('img');
-            var url = window.URL || window.webkitURL;
-            img.src = url.createObjectURL(this.response);
-            var show_thumbnail=document.getElementById("show_thumbnail");
-            show_thumbnail.appendChild(img);
+    $.ajax({url:new_url,success:function(response){
+        if(response.size!=233){
+            $('#show_thumbnail').append($('<img src="'+new_url+'" />'));
```
vbanos: You shouldn't write the image this way. Its better:
```
$('<img>').attr('src', new_url)
```

test/overview.spec.js
```diff
@@ -0,0 +1,16 @@
...
+        var captures = {
+            '2001': {
+                'text/html': 1,
+            },
+            '2002': {
+                'text/html': 4
+            }
+        }
+        expect(getTotal(captures)).to.equal(5);
+        //expect(5).to.equal(5);
```
vbanos: Drop commented code

scripts/overview.js
```diff
@@ -1,116 +1,82 @@
...
-        else{
-            document.getElementById("show_thumbnail").innerHTML="Thumbnail not found";
-        }
-    }
-    xhr.onerror=function(){
-        document.getElementById("show_thumbnail").innerHTML="Thumbnail not found";
-    };
-    xhr.ontimeout = function() {
-        document.getElementById("show_thumbnail").innerHTML="Please refresh the page...Time out!!";
-    }
```
vbanos: You have dropped the old code to handle error/timeout cases but you haven't put the relevant jQuery code to handle the same cases. Its also possible in jQuery to handle these errors, please look it up and add it as well.

vbanos:
General comment: please use `eslint` to fix code styling issues, in `overview.js`, there are too many of them...

abhidas17695:
@vbanos I addressed the issues you pointed out. Please review the PR.

scripts/overview.js
```diff
@@ -1,116 +1,80 @@
...
-    xhr.send(null);
+function get_details () {
+  var url = getUrlByParameter('url')
+  var new_url = 'https://archive.org/services/context/metadata?url=' + url
+  $.getJSON(new_url, function (response) {
+    var type = response.type
+    $('#details').text(type)
+    var captures = response.captures
+    var total = 0
+    total = getTotal(captures)
```
vbanos: Why don't you just write `var total = getTotal(captures)` ?

PR #206, opened 2018-10-24
Fixed bug regading alexa and doi conflict in context auto-update

anishsarangi:
The following bug has been fixed:-
When I select option "Show context in tabs", I pick Alexa & domaintools contexts, I enable auto-update-context.
When I visit one site, contexts appear correctly. When I visit another site, contexts refresh but instead of Alexa as the first context, I see the papers context


vbanos:
I see that the same problem exists when opening contexts in windows. Please fix that as well.
https://github.com/internetarchive/wayback-machine-chrome/blob/396dc641026613697bbb42dceb20bf461a34dbef/scripts/background.js#L886

anishsarangi:
Sorry, forgot to do that earlier. I have done the changes. Thank you @vbanos .

PR #247, opened 2018-11-11
Refactored chrome.storage.sync.get calls in background

abhidas17695:
@vbanos I refactored background.js such that chrome.storage.sync.get is not called multiple times. I also solved the problem Max described ( asynchronous nature of chrome tabs/windows opening). Please review this PR

### DIFF
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -431,7 +431,6 @@ function auto_save(tabId) {
 
 //function for opeing a particular context
 function openThatContext(temp, url, methodOfShowing) {
-  console.log("RECEIVED ", temp, url, methodOfShowing);
   return function () {
     return new Promise(function (resolve, reject) {
       var alexa_url = "https://archive.org/services/context/alexa?url=" + url;
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -23,49 +23,63 @@ var contexts = [
     htmlUrl: "https://archive.org/services/context/alexa?url=",
     tab: 0,
     window: 0,
-    tabContextName: 0
+    tabContextName: 0,
+    top: 0,
+    left: 0
   },
   {
     name: "domaintools",
     htmlUrl: chrome.runtime.getURL("domaintools.html") + "?url=",
     tab: 0,
     window: 0,
-    tabContextName: 0
+    tabContextName: 0,
+    top: 500,
+    left: 0
   },
   {
     name: "tweets",
     htmlUrl: 'https://twitter.com/search?q=',
     tab: 0,
     window: 0,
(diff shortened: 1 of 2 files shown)


scripts/background.js
```diff
@@ -173,235 +171,216 @@ chrome.webRequest.onCompleted.addListener(function(details) {
...
+          shouldOpen: event.tagcloud,
+          url: message.url
+        }
+      ];
       //If the Context is to be showed in tabs
-      if(event.show_context=="tab"){
-        if(tabId2==0 || tabId3==0 || tabId4==0 || tabId5==0 || tabId6==0 || tabId7==0 || tabId8==0 || tabId9 ==0){  //Checking if Tabs are not open already
+      if (event.show_context == "tab") {
+        if (tabId2 == 0 || tabId3 == 0 || tabId4 == 0 || tabId5 == 0 || tabId6 == 0 || tabId7 == 0 || tabId8 == 0 || tabId9 == 0) {  //Checking if Tabs are not open already
           //If not selected show-all option ,then check and open indivisually
```
vbanos: This comment is now irrelevant and should be removed.

vbanos:
A general comment that applies everywhere: When checking for equals, please use `===` instead of `==`.
When checking for not equals, please use `!==` instead of `==`.

scripts/background.js
```diff
@@ -13,31 +13,31 @@ var excluded_urls = [
...
-var windowIdSingle=0;
+var tabId2 = 0;
+var tabId3 = 0;
+var tabId4 = 0;
+var tabId5 = 0;
+var tabId6 = 0;
+var tabId7 = 0;
+var tabId8 = 0;
+var tabId9 = 0;
+var tabId10 = 0;
```
vbanos: We used to have 10 context windows but now we've got just 7.
The extra `tabId` and `windowId` vars need to be removed from everywhere (they are also present in some `if` conditions all over `background.js`.

scripts/background.js
```diff
@@ -173,235 +171,216 @@ chrome.webRequest.onCompleted.addListener(function(details) {
...
+          name: "similarweb",
+          shouldOpen: event.similarweb,
+          url: url
+        },
+        {
+          name: "tagcloud",
+          shouldOpen: event.tagcloud,
+          url: message.url
+        }
+      ];
```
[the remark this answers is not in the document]
abhidas17695: I am making changes the way you suggested @vbanos

abhidas17695:
@vbanos I have refactored `openThatContext()` to use the context objects. Please review the changes.

PR #256, opened 2018-12-03
Edited for firefox compatibility

abhidas17695:
@vbanos I edited background.js for compatibility with Firefox. Some bugs remain
1. If contexts are shown in tabs , the first context does not update if a new web page is opened.
2. Find cited books/papers doesn't work if single window chosen
3. 'Show me context' cannot be clicked on a new page if existing contexts are open for both tabs and windows . The new tabs , windows which open re-render infinitely.
None of the changes I made break any functionality for Chrome. I will be fixing the above mentioned bugs.

### DIFF
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -120,9 +120,9 @@ chrome.windows.onRemoved.addListener(function (id) {
   var index = contexts.findIndex(e => e.window === id);
   if (index >= 0) {
     contexts[index].window = 0;
-  }else if(windowIdtest === id){
+  } else if (windowIdtest === id) {
     windowIdtest = 0;
-  }else if(windowIdSingle === id){
+  } else if (windowIdSingle === id) {
     windowIdSingle = 0;
   }
 });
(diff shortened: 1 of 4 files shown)


scripts/background.js
```diff
@@ -373,13 +373,13 @@ chrome.tabs.onUpdated.addListener(function (tabId, info) {
           }
           if (event1.auto_update_context === true) {
             if (event1.show_context === "tab") {
-              if (contexts.findIndex(e => e.tab !== 0) >= 0 || windowIdtest !== 0) {
+              if ((contexts.findIndex(e => e.tab !== 0) >= 0 || windowIdtest !== 0)) {
```
MaxReinisch: The extra parenthesis shouldn't be necessary

abhidas17695: Thanks @MaxReinisch 
That is fixed now

abhidas17695:
@vbanos the extension now works completely in Firefox without any bugs.

PR #262, opened 2018-12-20
Single window refactor

abhidas17695:
@vbanos I have refactored singleWindow.js to use functions already defined in the scripts for each context. I couldn't figure out the method for using an iframe to display the tweets inside the single window. There are restrictions on executing inline scripts ( inside https://twitter.com/search?q=github.com ) in extensions. There are ways of by-passing these restrictions but I have to research them further. The extension works in both Chrome and Firefox.


vbanos:
I enable all contexts.
I don't enable to search for books or papers.
I visit https://en.wikipedia.org/wiki/Easter_Island and click show contexts in single window.
Problems:
1. First of all, I see "Cited Papers". Why? Its not a context. In addition, I haven't enabled to search for papers.
2. The names of the contexts in single window mode do not match exactly the names of the contexts when you view contexts in tabs. Please fix single window to have exactly the same names.

vbanos:
Problem:
I enable only the first two contexts.
I enable single window mode and click "Show contexts".
I see ALL contexts.

Regardless of the contexts I enable, everything is displayed in single window mode. Only enabled contexts should be displayed.

abhidas17695:
#1 I will edit singleWindow.js to use Domaintools
#2 Cited papers is shown because regardless of what contexts you choose to see , all contexts are shown in single window. I will hide the Cited Papers section on a non-Wikipedia page or if the option for seeing Cited papers is not chosen. And to answer why the titles of the sections do not match those of the contexts when shown in tabs, I have not edited those titles. The titles are exactly the same as what you would see on versions of the extension before I made the PR. This PR does not fix the issue you mentioned. I will make those changes.
#3 As mentioned in #3 , regardless of what options you select , all contexts are shown in single window mode. This isn't a bug in my PR but a bug in the existing code base. I will fix that.

@vbanos my point is that this PR isn't to fix any of the stuff you mentioned. This PR just refactors single window to use functions already defined. I agree with your first comment. But for comment 2 and 3 I must say that this PR isn't supposed to address them in the first place. Anyway I will fix those and make new commits.

abhidas17695:
@vbanos I addressed your first 2 comments in the latest commits. However your last comment is not addressed because the issue is totally not related to what this PR tries to achieve. I will soon make a new PR which will fix the issue you had mentioned.

scripts/singleWindow.js
```diff
@@ -1,191 +1,38 @@
...
+        }
+        var rl = xml.getElementsByTagName('RL');
+        if (rl.length > 0) {
+            for (var i = 0, len = rl.length; i < len && i < 5; i++) {
+                var title = rl[i].getAttribute('TITLE');
+                var href = rl[i].getAttribute('HREF');
+                var list_item = $("<li>");
+                var link = $("<a>").attr('href', "http://" + href).attr('target', '_blank').attr('class', 'rl-a').text(title.length > 18 ? title.substring(0, 15) + '...' : title);
+                list_item.append(link);
+                $("#alexa_list").append(list_item);
```
vbanos: This can be written much more efficiently:
```
$('#alexa_list').append(
  $('<li>').append(
    $("<a>").attr('href', "http://" + href).attr('target', '_blank').attr('class', 'rl-a').text(title.length > 18 ? title.substring(0, 15) + '...' : title)
  )
)
```

abhidas17695: Done !!

PR #270, opened 2019-01-04
Add AGPL license file

quassy:
As wayback-machine-chrome is licensed under GNU AGPL it should also include the corresponding license file (copied from wayback-machine-firefox)


vbanos:
Thank you!

Issue #274, opened 2019-01-08
What are the issues to which contributors can contribute?


GaurangTandon:
Hi,

I notice that [commits](https://github.com/internetarchive/wayback-machine-chrome/commits/master) are being made almost on a daily frequency to this repo. That means a lot of stuff has been done and has to be done. I feel I could have helped in several of these tasks, like, refactoring the JS files, adding donation buttons, etc. 

The [issues](https://github.com/internetarchive/wayback-machine-chrome/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) that are there currently are months old, and many of them have become obsolete ([example](https://github.com/internetarchive/wayback-machine-chrome/issues/6)) 

Could the maintainer please create a few issues describing what are the tasks they need help with next, to which a contributor like me can contribute?

anishsarangi:
You can start with refactoring the JS files.
Thank you.

GaurangTandon:
Hi @anishsarangi , thanks for your reply. 

I have a question: Do you use a style guide, like `.editorconfig`, or `.prettierrc`? There are chances that our tab style/quote style/function style/etc. might not match, and I would be glad to have consistency throughout the repo. If you want should I create a basic `.prettierrc`/`.editorconfig` file? Once that is done, and once the maintainers confirm if the style specifications are ok with them, I can also run all the js files one by one through this style guide, so they all are consistent, and create a PR for the same. Thoughts?

mkfeuhrer:
@anishsarangi @vbanos could you provide with some future features or issues on which contributions can be made. Can't find any relevant issues. If there is could you point me to one. Thanks.

daveloyall:
The two most recent committers are @MaxReinisch and @vbanos.  Hello!

A user on IRC asked:

> Who do I contact regarding contributions to the Wayback Machine Chrome Extension?                                                                                                                                                                            

I replied:

> I am not part of the archive.org team.  But, I can tell you that this source code repository is active:  https://github.com/internetarchive/wayback-machine-chrome
> There do seem to be some issues and pull requests there.

The user replied:

> There are issues and PRs, yeah. But, there is no response from the people who are actively committing.                                                                                                                                                    

Maybe it is time for some kind of contributors' guide document?  Or for a review of tickets?  Or to grant some additional person permission to merge PRs or close issues?

Keep up the good work.  Cheers!

radicalswetank:
[...]

PR #278, opened 2019-01-31
refactored popup.js for better extensibility

GaurangTandon:


GaurangTandon:
Hi, as given in the advice in [this issue](https://github.com/internetarchive/wayback-machine-chrome/issues/274), I have attempted to refactor the popup.js file. Please let me know what else to add to the PR, so that it may be merged. Thanks!

GaurangTandon:
@anishsarangi Could you please review this PR? It's been here for a quite long time here. I'd be happy to extensively refactor other `.js` files as well.

vbanos:
Thank you for your effort but this PR does not improve `popup.js`.

For instance, when I see the following, it is really confusing...
```
function recent_capture() {
  sendOpenUrlMessage(2, 'recent');
}
```
The reader has to read `sendOpenUrlMessage` to understand that `2` is used to make path `/2/` and there is no comment explaining that is the purpose of this path.

The same stands for how you made `get_clean_url`.

I'm sorry but we cannot accept it.

PR #296, opened 2019-03-26
Refactor index.html file

akkywadhwa:


vbanos:
I'm sorry but there are far better ways to do this without introducing so many new CSS classes.

E.g. we use bootstrap 3 and it already has classes to hide elements, we should reuse those. https://getbootstrap.com/docs/3.4/css/#helper-classes-show-hide

Also, you could add CSS attributes to existing element ids without defining new ones.

PR #319, opened 2019-04-03
fixing issue/314

Zerichen:


MaxReinisch:
Looks great.  Just had the though, is the function `get_amazonbooks` used anywhere? Is it a necessary function to have? If not, we should probably just delete amazon.js

Zerichen:
You are right! We can delete it right now I think.

anishsarangi:
@MaxReinisch I am sorry, but this creates another bug. If you open https://www.amazon.com/Easter-Island-Novel-Jennifer-Vanderbes/dp/0385336748 and then open https://www.amazon.com/Finding-Public-Domain-Copyright-Management/dp/1607853736/ , when I tried to go back to the first URL and click "Read book now", then it opens the archive.org copy of the second URL book.
Can you try this ?

Zerichen:
Hi @anishsarangi, sorry about that! I think I can fix this by calling ``` chrome.sync.set()```inside the ```chrome.tabs.update()``` . In this way, the sync data can get updated when a tab's URL varies while not making a new request. Let me see if it will work.

MaxReinisch:
good catch @anishsarangi.  I am actually seeing a slightly different issue than what you described.  For me I go to the first book and it works as intended, but then it doesn't work for a new tab with a second book.  Can you please share the urls that you used while testing this?

@Zerichen You shouldn't have to save the book info into memory.  If we make the request in popup.js in the function borrow_books, we can then use messaging to change the icon badge.  This is more similar to how auto_save works compared to how it used to work.  So in borrow_books, it should check the amazon setting, verify that it is a url from amazon books, then make the request, and if the response is positive, add the button and send message to background script to change the badge.

Zerichen:
@MaxReinisch Get it.

anishsarangi:
@MaxReinisch Please go to https://www.amazon.com/Easter-Island-Novel-Jennifer-Vanderbes/dp/0385336748 and click on "read book now". Then you go to https://www.amazon.com/Finding-Public-Domain-Copyright-Management/dp/1607853736/ and click on "read book now". Then please open the tab containing first URL, and click on "read book now", you can see the book for the second URL.

anishsarangi:
@MaxReinisch popup.js donesn't execute untill you click on the extension icon.

MaxReinisch:
Interesting, that makes this problem a bit harder.  I guess the plan should be to implement the request in background.js in the tab update listener like how it is now, but instead of using the chrome storage sync, you can try sending a message to popup.js

Zerichen:
@MaxReinisch @anishsarangi Maybe we could store the tabId as the key rather than ''details_url'' so that a new request URL won't overwrite the previous one?

MaxReinisch:
@Zerichen then we would be flooding the memory

Zerichen:
@MaxReinisch Get it. I will try to fix this later on.

MaxReinisch:
I am going to revert this PR as to not have the bug in production code.  Please open a new PR from the issue/314 branch

Zerichen:
@MaxReinisch Sure! I will fix it asap.

PR #320, opened 2019-04-03
Fix 315: wikipedia books request now in background.js

Zerichen:


Zerichen:
Hi @MaxReinisch, in this PR, I rewrite the getWikipediaBooks function and make it return a promise for both addCitations() and populateBook() to handle. And a message listener is being added to the background.js file as well.

MaxReinisch:
Everything looks great.  Not sure why you removed amazon.js in this branch though.  That should be done in the branch addressing that issue.  Revert that change and it looks good. 

Also, I like how you have the issue number in your PR, however it also is helpful if you include some information about the issue, like "Fix 315: wikipedia books request now in background.js"

Zerichen:
@MaxReinisch I have reverted the change.

PR #321, opened 2019-04-03
Fixed improper usage of "isNotExcludedUrl" 

cheevu-RS:
The function isNotExcludedUrl returns true if the URL is not excluded, false if it is excluded. In function **auto_save** , !isNotExcludedUrl is being used, when it is to be just isNotExcludedUrl.
This was causing a bug wherein BadgeText "S" was not being added to the icon, when the page had not been archived.


anishsarangi:
Thank you @cheevu-RS.

PR #322, opened 2019-04-03
Fixed #316

### LINKED ISSUE #316: Wikipedia papers isn't working right
On the page: https://en.wikipedia.org/wiki/Nan_Madol, using the "Find cited papers" feature has a UI bug.  I am also wondering if this should make the API call in background.js like how wikipedia.js should be  #315

cheevu-RS:
Fixed UI bug in find cited papers.

### DIFF
--- a/css/doi.css
+++ b/css/doi.css
@@ -39,6 +39,7 @@
 
 .bottom_details{
   justify-content: flex-end;
+  text-align: center;
   margin: 10px auto;
 }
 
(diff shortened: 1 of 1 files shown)


MaxReinisch:
Please use the wikipedia books feature as a reference for the style.  


image

cheevu-RS:
Is this fine?
The class name was not matching in the doi.css.
Screenshot from 2019-04-04 11-22-26

PR #326, opened 2019-04-04
Fixing Issue/314: amazon books now will only make one request

Zerichen:
@MaxReinisch @anishsarangi 
**Explanation of the previous bug:** 
Since Chrome browser will cache your page, once we switch back to the tabs which we have already requested, chrome will not update the tab hence ```fetch``` will not execute. In this situation, the url stored in the chrome storage will not be updated as well so we will see that the extension is fetching the previous book.

Sorry about the fact that I am still using chrome storage to implement it this time because it is the easiest way to get shared data across all extension files in my opinion. Besides, I will only store two key-value pairs so memory throttling might not be a problem here.  

**Current solution:** 
For each request from ```background.js```, I store two key-value pairs, which are the tab url and the fetched archive url. Once the badgeText has B in it, then we can check if the current tab url is what we have stored just now. If so, then it means that the current tab is not using cache and our archive url is valid to use. Otherwise, it means that this page is using cache hence not making a new fetch request to update the chrome storage. So this time we can make a request from ```popup.js``` to get the new archive url. (In my opinion, this request is not a redundant request since we can not store the archive url for each request)

Let me know if there is any bug this time! I have tested for switching among 4 pages and it works fine. But maybe there is some other situations I haven't think about yet.

### DIFF
--- a/scripts/popup.js
+++ b/scripts/popup.js
@@ -209,16 +209,13 @@ function borrow_books() {
           let stored_url = result.tab_url
           let detail_url = result.detail_url
           // Checking if the tab url is the same as the last stored one
-          console.log(stored_url, url, detail_url)
           if (stored_url === url) {
-            console.log('no request making in popup.js')
             // if so, then we can use the previously fetched url
             $('#borrow_books_tr').css({ 'display': 'block' }).click(function () {
               chrome.tabs.create({ url: detail_url })
             })
           } else {
             // if not, we can then fetch it again
-            console.log('making request in popup.js', url)
             get_amazonbooks(url).then(response => {
               if (response['metadata'] && response['metadata']['identifier-access']) {
                 let details_url = response['metadata']['identifier-access']
--- a/scripts/popup.js
+++ b/scripts/popup.js
@@ -205,17 +205,20 @@ function borrow_books() {
     tabId = tabs[0].id
     chrome.browserAction.getBadgeText({ tabId: tabId }, function (result) {
       if (result.includes('B') && url.includes('www.amazon') && url.includes('/dp/')) {
-        chrome.storage.sync.get(['bg_url'], function (result) { 
-          let stored_url = result.bg_url
+        chrome.storage.sync.get(['tab_url', 'detail_url'], function (result) { 
+          let stored_url = result.tab_url
           let detail_url = result.detail_url
           // Checking if the tab url is the same as the last stored one
+          console.log(stored_url, url, detail_url)
           if (stored_url === url) {
+            console.log('no request making in popup.js')
             // if so, then we can use the previously fetched url
             $('#borrow_books_tr').css({ 'display': 'block' }).click(function () {
               chrome.tabs.create({ url: detail_url })
             })
           } else {
             // if not, we can then fetch it again
+            console.log('making request in popup.js', url)
             get_amazonbooks(url).then(response => {
               if (response['metadata'] && response['metadata']['identifier-access']) {
                 let details_url = response['metadata']['identifier-access']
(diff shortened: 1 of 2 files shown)


anishsarangi:
@Zerichen This only useful for the FIRST url for not to again request the API. But if we are opening another URL it is making the request twice.

Zerichen:
@anishsarangi Do you mean that when you go back to a cached tab, ```popup.js``` will make another request? From my consideration, this second request is not redundant since previously we are avoiding the request sending from both ```popup.js``` and ```background.js``` if we open a book link from amazon.

anishsarangi:
YES, That's what I am saying. It works in that case. 
Let me explain :
If you will go to https://www.amazon.com/Finding-Public-Domain-Copyright-Management/dp/1607853736/ you can see that only request going from background.js . Then if you will go to https://www.amazon.com/Easter-Island-Novel-Jennifer-Vanderbes/dp/0385336748,  you can also see that request is going only from background.js. But If I again go to the tab containing the first URL and click on read now button , in this case the request will occur in both background.js and popup.js .
Did you get it ?

scripts/popup.js
```diff
@@ -205,14 +205,30 @@ function borrow_books() {
...
     chrome.browserAction.getBadgeText({ tabId: tabId }, function (result) {
       if (result.includes('B') && url.includes('www.amazon') && url.includes('/dp/')) {
-        get_amazonbooks(url).then(response => {
-          if (response['metadata'] && response['metadata']['identifier-access']) {
-            let details_url = response['metadata']['identifier-access']
+        chrome.storage.sync.get(['tab_url', 'detail_url'], function (result) { 
+          let stored_url = result.tab_url
+          let detail_url = result.detail_url
+          // Checking if the tab url is the same as the last stored one
+          console.log(stored_url, url, detail_url)
```
anishsarangi: Please try not to include these statements in your final code.

Zerichen: Yes, I am adding these for your debugging purpose. I will delete this once you have approved my PR

Zerichen:
> YES, That's what I am saying. It works in that case.
> Let me explain :
> If you will go to https://www.amazon.com/Finding-Public-Domain-Copyright-Management/dp/1607853736/ you can see that only request going from background.js . Then if you will go to https://www.amazon.com/Easter-Island-Novel-Jennifer-Vanderbes/dp/0385336748, you can also see that request is going only from background.js. But If I again go to the tab containing the first URL and click on read now button , in this case the request will occur in both background.js and popup.js .
> Did you get it ?

Sorry I tried it but my console was showing that the background.js has only made two request in total in that  case. I am wondering how you go back to the tab? For me, I just simply click on the tab containing the first url which is using cache hence it is unreasonable for the tab to be updated?

anishsarangi:
I tried opening  https://www.amazon.com/Finding-Public-Domain-Copyright-Management/dp/1607853736/ in a new tab and saw that only request going from background.js . Then I opened another URL https://www.amazon.com/Easter-Island-Novel-Jennifer-Vanderbes/dp/0385336748  in an **another** tab, also saw that request is going only from background.js. But then I again went to the tab containing the first URL and click on read now button , in this case the request occured in both background.js and popup.js . (saw both consoles of background.js and popup.js)
Can you try it?

Zerichen:
Yes, I tried the same thing but my background console did not make the third request. Really weird tho.

anishsarangi:
Ahh.. After reinstalling the extension, it is showing me just 2 requests.
Good work @Zerichen. I think, this method is a correct one :)

Zerichen:
Haha debugging the extension is kind of annoying. I will notify you when I deleted the testing code and you can merge it then.👍

MaxReinisch:
It gets easier to debug over time. You can set breakpoints in the chrome js console and debugger though.

Zerichen:
Never used that before but thanks for pointing out👌

Zerichen:
@anishsarangi @MaxReinisch I have removed the test code and I think this branch is ready to be merged right now!

sr6033:
nice work @Zerichen

Zerichen:
Thanks @sr6033

PR #328, opened 2019-04-04
Bug fixed regarding the S badge not turning into a checkmark in auto-save feature

anishsarangi:
@vbanos @MaxReinisch This is a fix to the bug "the S badge not turning into a checkmark in auto-save feature"

### DIFF
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -478,11 +478,12 @@ function auto_save(tabId, url) {
         })
       },
       function () {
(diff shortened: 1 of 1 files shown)


vbanos:
You don't explain what was the problem in the previous logic and how you fix it with what you do in this PR.
Please try to do that.

anishsarangi:
@vbanos The main logic was when a page is not in the archive and we are going to show user "S" and after archiving it should show a check-mark. This was not happening. This PR is a fix to that.

PR #335, opened 2019-04-06
Optimized tag's weight find concept and fix tag-cloud issue in single page view

sr6033:
Removed the function findWeightOf() as it is no longer needed now. Now the complexity has reduced from O(n x m) to O(m).

### DIFF
--- a/scripts/singleWindow.js
+++ b/scripts/singleWindow.js
@@ -170,9 +170,9 @@ function percentEncode (str) {
     return '%' + character.charCodeAt(0).toString(16)
   })
 };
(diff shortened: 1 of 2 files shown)


sr6033:
@MaxReinisch 
Can you check this out?

sr6033:
The 2nd commit fixes the issue #336

MaxReinisch:
Good job 👍

anishsarangi:
Good work @sr6033 . Also please remember to make two separate PRs for addressing two separate issues.
Thank you.

sr6033:
Okay, @anishsarangi. I will keep that in mind from now on.

Issue #339, opened 2019-04-09
Add Support to check if any Contexts are selected before sending 'showall' message


thisisayush:
The `showall` message is sent even when there are no contexts selected or initialized. This results in poor UX as first-time users may wait thinking something will happen.
What we can do is add a check before sending the message and showing a message to user if no contexts are selected.

sr6033:
I have included that in my new UI already.

sr6033:
This is a nice addition to the UX for new users. I was adding that in the experimental UI as well. Good work @thisisayush 

You can also integrate this into your new UI, I think.

MaxReinisch:
@thisisayush is right, @sr6033.  It is better to make these changes separately from the UI changes.  I understand that this change does fall under the category of UX, it is still more of a bug fix than a design change.

sr6033:
Yup. Also, what @MaxReinisch suggested in the PR can be looked into after some thoughts.

PR #340, opened 2019-04-09
(add) Context Selection Check before sending showall signal

thisisayush:
Refer #339 

#### Normal
image

#### *OnClick* if context is not available
image

This reverts back to original within one second.

### DIFF
--- a/scripts/popup.js
+++ b/scripts/popup.js
@@ -196,7 +196,29 @@ function settings() {
 
 function show_all_screens() {
   var url = get_clean_url()
-  chrome.runtime.sendMessage({ message: 'showall', url: url })
+  contexts = ['alexa', 'domaintools', 'tweets', 'wbmsummary', 'annotations', 'tagcloud']
+  var is_contexts_available = false
+  chrome.storage.sync.get(contexts, function (event) {
+    for(let x in contexts){
+      let context = event[contexts[x]]
+      if (context){
+        is_contexts_available = true
+        break
+      }
+    }
+    if(is_contexts_available)
+      chrome.runtime.sendMessage({ message: 'showall', url: url })
+    else {
+      let context_button = $("#context-screen")
+      context_button.attr('disabled', 'disabled')
+      let o_html = context_button.html()
+      context_button.html("No Context(s) Selected")
+      setTimeout(function(e){
+        context_button.html(o_html)
+        context_button.removeAttr('disabled')
+      }, 1000)
+    }
+  });
 }
 
 function borrow_books() {


MaxReinisch:
Wouldn't it make more sense to check if any of the settings are true (logical or) and if not, just hide the button entirely?

thisisayush:
I initially thought the same, but then the user will probably never know about the feature, imo.

sr6033:
Let's think of a workaround for it. Something that will let the user know and also won't show the button unless checked.

MaxReinisch:
How about leaving the button text the same, but graying it out and adding an explanation popup/title on hover.  Something like "Enable context features in settings"

thisisayush:
@MaxReinisch That sounds cool. I was thinking that instead of Showing this, How about we create a startup guide that launches before the agreement, that will involve configuring the Extension (if not already) with Info on what each option mean and then we show up the agreement?

In this way, the user will also get familiar with extension and we'll have contexts ready. So if the user has not selected any context, hiding the button will now make sense.

sr6033:
@thisisayush that can be a workaround but won't that be a bit heavy in terms of UX as it may bore the user to start configuring so much in the first run itself.

thisisayush:
I don't think that would be heavy, plus we'll be requiring user to do the minimum config, I personally came to know about what the context button does after some time when I randomly clicked those checkboxes in the settings

tanweer919:
@thisisayush How about adding a dropdown to manually select the feature the user wants to run. Because right now, on clicking the button all the selected feature pops out at once. This may hamper user experience.

sr6033:
Drop down will become a big list....and adding such a big drop down in an extension is generally not recommended.

tanweer919:
And how about, when the context button is clicked, we show a popup or a modal to select the context feature that the user wants to run. Because executing the features all at once as happening right now may not be very user friendly.

sr6033:
@tanweer919 even now the user selects what features to be shown. But I understand what you mean by that and that can be a good idea for UX. I will think about it and comment a solution.

tanweer919:
Yes the user can still select the featues to be shown, but for this they have to go within the settings option. I think we can make it simpler.

MaxReinisch:
I agree that the current way the feature works is not a great user experience.  However, it is the way the idea was conceived, and we've been discussing how to improve the feature overall.  While this discussion has been great, I'd like to see something be done with the UI now because the overall issue of "Having a button that does nothing when no settings are on" is really bad UI.  

I will open an issue regarding overall context ui ideas.  But for now, I'd like to see the "gray out + suggestion to turn on context features in the settings" idea committed @tanweer919

thisisayush:
Hi @MaxReinisch, sorry for the late response, I've my exams going on from 15th Apr all the way till 15th May, resulting in low contributions. As suggested, I'll grey out the button and write a text below the button suggesting user to select contexts? Sounds cool?

anishsarangi:
Thank you for this PR. But now we are moving in different direction regarding this UI. We can close this PR. If you feel any other improvements need to be done, please make a PR.

PR #344, opened 2019-04-09
booklist.js and domaintools.js are being optimized

Zerichen:
Since using jQuery to create objects is more expensive than using native javascript api, especially in a for loop, I have replaced many jQuery implementations with the plain javascript code. Besides, reducing the number of append to an existing DOM node will greatly reduce the time of reflow as well. For normal use cases, the running time of function has been optimized from 2.9ms to 0.8ms. The difference will be more obvious when DOM is rendering a large number of objects.
There are still some files left that can be optimized in a similar way, I will try to finish them in the future.

### DIFF
--- a/scripts/booklist.js
+++ b/scripts/booklist.js
@@ -1,9 +1,12 @@
 function populateBooks(url) {
   // Gets the data for each book on the wikipedia url
   getWikipediaBooks(url).then(data=>{
-    $(".loader").hide();
-    // No need to check the validation of data since promise is already being resolved
-    console.time()
+    const loaders = document.getElementsByClassName('loader');
+    for (const i of loaders) {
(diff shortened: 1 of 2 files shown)


MaxReinisch:
The issue I see with this is that the requests are the biggest bottleneck.  Who cares about shaving 2ms off when the request will be on the scale of seconds.  So the negative aspects of using JQuery are pretty minimal.  That being said, JQuery is great at making code cleaner and easier to read.  We are not trying to go back to vanilla JS.

Zerichen:
@MaxReinisch Thank you for the comment on my rash PR. Sorry about that! I will pay more attention to the syntax consistency of our codebase and try to approach the optimization in a more effective way next time!

PR #349, opened 2019-04-16
Fixed arrow key issue within dropdown

tanweer919:
Create this PR with a branch other than master


tanweer919:
@MaxReinisch done

Issue #338, opened 2019-04-08
Maintaining CodeBase Consistency


thisisayush:
Currently in most of the Javascript files the identifiers (function names, variables, etc) don't follow a standard naming convention, some follow the underscore separated names i.e. `borrow_books` while some follow the CamelCase pattern e.g. makeModal`

Another issue to address in this is Inconsistent Indentation across files. Some files use 4 space Indentation while some are using 2 space Indentation.

There should be a proper naming convention and indentation guidelines and should be added in the contribution guidelines that @MaxReinisch is working on (#325 ).

sr6033:
**ESLint** is already being used in the project. Only a proper guideline is to be mentioned for developers contributing to the project.

tanweer919:
Callback definition is also inconsistent. In some places es6 arrow functions are used while in othe places, traditional javascript functions are used.

sr6033:
Transferring the codebase to ES6 is a different issue altogether, and we need to work on it if is approved by high up @tanweer919.
@thisisayush I agree with your conventions. Seems good and consistent.

cheevu-RS:
@sr6033 I guess, they do want to move on to ES6, from Issue #307's response.

YashKumarVerma:
> @sr6033 Yes ESLint is being used, however, there is a lot of inconsistency in naming conventions. Adding it to Guidelines is something that we can use to set standards to prevent such inconsistencies in the future, however, currently existing codebase needs to be made consistent.
> 
> I suggest the following convention after observing the code (for JS):
> 
> * Using camelCase for functions/methods
> * Using lowecase_underscore for a variable name
> * Using UPPERCASE for Constants

How about using camelCase for the variable names as well ? Won't that be more consistent?

thisisayush:
It'll be, but I suggested the convention according to what was being used majorly in the code.

YashKumarVerma:
@thisisayush  can you have a look at #424.

Also, there's a mixture of both notations in the code. So I was a bit confused regarding it.

BisariaUtkarsh:
> @sr6033 Yes ESLint is being used, however, there is a lot of inconsistency in naming conventions. Adding it to Guidelines is something that we can use to set standards to prevent such inconsistencies in the future, however, currently existing codebase needs to be made consistent.
> 
> I suggest the following convention after observing the code (for JS):
> 
> * Using camelCase for functions/methods
> * Using lowecase_underscore for a variable name
[...]

Issue #307, opened 2019-03-29
Converting to ES6


akkywadhwa:
I think we can use ES6 arrow functions in various javascript files. Should I make a PR for it?

anishsarangi:
YES. Please !!
Thank you.

darkshredder:
I would like to work on this issue.

PR #351, opened 2019-04-17
Converted to Arrow functions in citations.js

radicalswetank:
#307 #338
I have converted some existing functions to **arrow functions** in **citations.js** file as a result of the conversation of the above-referenced issues. If this is a desired change and my PR is accepted. I will put more PR's for converting the existing codebase to ES6.

### DIFF
--- a/scripts/citations.js
+++ b/scripts/citations.js
@@ -15,7 +15,7 @@
 //   findCitations()
 // })
 
-function findCitations () {
+var findCitations = () => {
   let candidates = $("a[href^='#_ftnref']").parent()
   for (let i = 0; i < candidates.length; i++) {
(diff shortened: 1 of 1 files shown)


Zerichen:
Good work. Since we are trying to follow the ES6 standard, I think it could be even better to use ```const / let``` instead of ```var```?

sr6033:
`const` and `let` are a good suggestion @Zerichen

MaxReinisch:
I think you're missing the point of arrow notation.  It creates an unnamed function, which of course can be named using the code you wrote.  But unnamed functions are used commonly as callbacks in javascript.  Writing all functions in this way don't really have any benefit, and is simply harder to read than using the `function` keyword.

PR #352, opened 2019-04-17
Removed redundant CSS

radicalswetank:
`.rec_results` class was redundant So I have removed the class

### DIFF
--- a/css/booklist.css
+++ b/css/booklist.css
@@ -59,12 +59,6 @@ body{
 .container{
   padding: 15px;
 }
-.rec_results{
-  background-color: white;
-  padding-left: 15px;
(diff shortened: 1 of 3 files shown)


thisisayush:
@radicalswetank The classes you removed were for specific pages, i.e. they seem redundant but they were used in a page only once. Please check if the classe is still available to the page, the CSS file was a part of?

radicalswetank:
@thisisayush I have already checked it and I have also checked old commits for the same!!!

anishsarangi:
Thank you for this PR. Please resolve the conflicts so that we can merge it ! If you feel any other improvements need to be done, please make a PR.

PR #358, opened 2019-04-20
Refurbished UI : Concise Extension

Ritvyk:
Review the changes , new 'Square menu' & 'DropDownArrow' icon is to be uploaded to the images.
Updated the previous existing UI with nested Drop Down menu.
New Concise UI will give a nice user experience as one would not be happy to see too many options on a extension.

wayback

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -5,6 +5,35 @@ li{
     margin:0px 0px 0px -18px;
     padding: 0px;
 }
+.dropdown,.dropdown2{
+    display: inline-block;
+    position: relative;
+    width: auto;
+    height: auto;
+}
+.dropdown2{
+    display: inline-block;
+    position: relative;
+    width:auto;
+    height: auto;
+}
+.menu_content2{
+    display: none;
+    height:auto;
+    width: 165px;
+}
+.menu_content{
+    display:none;
+}
+.dropdown:hover{
+    transform: scale(1.02);
+}
+.dropdown:hover .menu_content{
+    display: block;
+}
+.dropdown2:hover .menu_content2{
+    display: block;
+}
 #horizontal_line{
     margin: 0px; 
     padding: 0px; 
--- a/index.html
+++ b/index.html
@@ -29,16 +29,23 @@
         </form>
         <button type="button" class="btn btn-sm btn-block" id="save_now" tabindex='-1'>Save
           Page Now</button>
-        <button type="button" class="btn btn-sm captures btn-block" id="recent_capture" tabindex='-1'>Load Recent
-          Version</button>
-        <button type="button" class="btn btn-sm captures btn-block" id="first_capture" tabindex='-1'>Load First Version</button>
-        <button type="button" class="btn btn-sm captures btn-block" id="overview" tabindex='-1'>Browse All Versions</button>
-        <button type="button" class="btn btn-sm extra_feature btn-block" id="make_modal" tabindex='-1'>Visualize Site
-          Map</button>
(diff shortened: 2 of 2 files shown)


Ritvyk:
@Zerichen  @thisisayush  @tanweer99  @sr6033  If you can have a look over it and review it.

Ritvyk:
@MaxReinisch Please Have a look upon it , The 'MenuSquare' icon and "dropDownArrow' icon is to be uploaded and any feedback would be expected.

sr6033:
Looks good but it makes the user go through too many clicks. Can you think of a better approach to concise the look?

Zerichen:
@sr6033 Agree, I think the popup page is already very small therefore there is no need to fold those options, but it's still a good attempt!

Ritvyk:
@sr6033 no need of clicks actually , it drops down , by hovering over it ! By the way thanks for the feedback @sr0633 @zerichen

Ritvyk:
@sr6033 and @zerichen what I Think is putting too much options on one extension isn't going to be liked by the users because it may get them confuse , instead necessary options are on the top of drop down and less ones are at down. Just thinking from a users perspective.

Zerichen:
@Ritvyk I think as long as the buttons are not disturbing for our users, it is fine to have multiple buttons in a single view.

sr6033:
As per user perspective, this should be the order of importance, I suppose (taking into consideration the frequent use of a feature):

| Order       | Feature                                           |
|----------|------------------------------------------------------------|
| 1.     | **Save page now**                                              |
| 2.    | **Load recent version**, **Load first version**, Browse all versions |
| 3.    | Show me context                                            |
| 4.    | Visualize site map, Tweets about this page                  |

Ritvyk:
@sr6033  Thanks Buddy for the feedback will take it into consideration.

Ritvyk:
@Zerichen @sr6033 @MaxReinisch  @thisisayush buddies i was thinking to add a voice command feature :
say : Save page as "<Name>" and it would automatically save the page in the archive , i am working on it , but it will be good if i discuss this with you guys !

MaxReinisch:
> As per user perspective, this should be the order of importance, I suppose (taking into consideration the frequent use of a feature):
> 
> Order	Feature
> 1.	**Save page now**
> 2.	**Load recent version**, **Load first version**, Browse all versions
> 3.	Show me context
> 4.	Visualize site map, Tweets about this page

This is fairly accurate, but don't forget other features like amazon books and wikipedia citations, etc.  Also, just as far as products go, some people will likely only use the extension for 1 or 2 features, like spn and recent version, or wikipedia books and papers.  But I agree with the point that displaying several buttons is not the same as displaying too many buttons, and it is more cumbersome to have to guide your mouse over buttons to get to the option you want.  

I'd categorize the features as "Wayback features", "Context features", and "Context features".  Twitter falls more in context feature, as it actually is a context feature that has its own separate button, and Site map I'd argue is a WB feature.  Maybe the best way to display the UI concisely would be to have 3 drop downs that you can open to access each feature.  When the extension popup is closed, it preserves which tables were left open.  That way, commonly used features remain accessible, and unused features remain hidden.

Ritvyk:
@MaxReinisch I think that's fair enough , will work on that !

Ritvyk:
@MaxReinisch please also look upon #365 and show your views over it !

sr6033:
@MaxReinisch That will be a nice way to categorize the features.

anishsarangi:
Thank you for this PR. But now we are moving in different direction regarding this UI. We can close this PR. If you feel any other improvements need to be done, please make a PR.

Issue #348, opened 2019-04-16
Ideas for better "Show me Context" UX.


MaxReinisch:
There was a great discussion over at #340 about how to best improve the UX of this feature, as it has been a problem for a while.  Please continue that conversation here if you'd like.

sr6033:
@MaxReinisch I am working on an idea for that. Will show the UI soon.

Ritvyk:
I was thinking to add voice command feature.
**say -> "save this page as <name>" and the page will be saved in the archive.**
we could add more commands its just the matter of discussion over it.
Would you like to approve this idea or discuss over it..
_i have also created the prototype for this functionality._

anishsarangi:
This thing is already implemented !! We can close this Issue !!

PR #361, opened 2019-04-24
Context options refurbish

sr6033:
Now the user can select the context options from the extension itself. No need to open settings page every time to change options. This leads to better and easy use of the feature by a user. They can easily add or remove context options when they want to see them. 
Issue: #348

image
image

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,3 +1,3 @@
 {
-    "extends": "standard"
+    "extends": "eslint:recommended"
 }
--- a/css/index.css
+++ b/css/index.css
@@ -5,28 +5,80 @@ li{
     margin:0px 0px 0px -18px;
     padding: 0px;
 }
+
+.show_context_options {
+    display: block !important;
+    background: #F4F4F4;
+    text-align: left;
+    padding-left: 10px;
+    padding-right: 10px;
+    padding-top: 5px;
+    padding-bottom: 1px;
+}
+
+.show_context_options button {
+    margin-bottom: 5px !important;
+}
+
+.show_context_options p {
+    margin: 0;
+}
+
+.show_context_options label {
+    font-weight: normal;
+}
+
+.show_context_options span {
+    margin-left: 5px;
+}
+
+.hide_context_options {
+    display: none;
+}
+
+.icon_dropdown {
+    position: absolute;
+    right: 12px;
+    top: 11px;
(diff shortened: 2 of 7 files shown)


sr6033:
@MaxReinisch 
Can you have a look and see if this is good?

Zerichen:
I also came up with a similar idea after seeing the discussion about issue #348 haha. In this way, the user can quickly change the context options based on their needs. Anyway, Good job!

sr6033:
Yup @Zerichen. Ease of use.

MaxReinisch:
I like it.  I am going to ping @markjgraham for this to see what he thinks.

MaxReinisch:
@sr6033 after using this some more, I realized a couple of things.  

1) This breaks the settings documentation.  If you can get the tooltips on the popup, that would be great.  If not, then delete the code that is trying to do this rather then let it throw errors.  
2) The drop down menu is a little jarring.  It would look fantastic with a drop down animation.  
3) The popup is now getting a little long.  I don't know how avoidable this is, and probably the best way to fix this would be condensing the WBM features into a single row in a different PR.  That being said, I do think that in this PR, the new context button can take up less room on average, while also requiring fewer clicks to use.  I think if the button functioned the way it used to, but there was a carrot button to display the context settings next to it, that would mean it is not necessary to open the context setting drop down every time you use the feature.  I hope I'm explaining this concept clearly, but if not, please ask for clarification.

sr6033:
@MaxReinisch I understood your first 2 points and fix them in the next PR. I would like you to help me out in understanding the 3rd point a bit more clearly.

MaxReinisch:
Instead of putting the "Show me context" button at the bottom of the context settings, put it above the toggle dropdown button.  

Then instead of having 2 rows of buttons, condense it to 1 row, with a dropdown button with the carrot (v) next to the "Show me context button."  If you are using bootstrap grid layouts, maybe make the "show me context" size 8 or 9 and the dropdown button size 3 or 4 (so long as they add up to 12 )

sr6033:
@MaxReinisch I got you now. I will send in a PR with the fix. 
And about this - 
> The popup is now getting a little long.
I will have to think about it on how it can be condensed in the present UI. Although, this can be easily accomplished in my newer Refurbished UI as the design supports collapsing of categories.

PR #365, opened 2019-04-28
Voice Input Feature

Ritvyk:
I was willing to take into account a voice command feature(Working on it !)
**say -> "save page as 'name' " 
execution -> automatically saving to archive**
currently this PR allows the user to say the title for the archive page that will be automatically converted into text.
NOTE: As input takes a valid url , https:// will be automatically appended.
waybackmachine

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -29,7 +29,14 @@ li{
     justify-content: center; 
     text-decoration: none;
 }
-
+i#MIC{
+    position: relative;
+    left:90px;
+    bottom:30px;
+}
+i#MIC:hover{
+    transform:scale(1.2);
+}
 button {
   display: block;
   margin-bottom: 10px !important;
--- a/index.html
+++ b/index.html
@@ -19,6 +19,7 @@
             <div class="bootstrap-typeahead-input" tabindex="-1" style="outline: none; position: relative;">
               <input id="search_input" type="text" placeholder="Search" value="" autocomplete="off" class="form-control has-aux bootstrap-typeahead-input-main has-selection"
                 style="background-color: transparent; display: block; position: relative; z-index: 1;">
+              <i onclick= "startConverting()" id='MIC'class="fa fa-microphone" title="Say 'Title' for archive Page."></i>
               <input type="text" tabindex="-1" value="" class="form-control has-aux bootstrap-typeahead-input-hint"
                 style="border-color: transparent; bottom: 0px; box-shadow: none; color: rgb(165, 165, 165); display: block; position: absolute; top: 0px; width: 100%; z-index: 0;">
             </div>
--- a/scripts/VoiceRec.js
+++ b/scripts/VoiceRec.js
@@ -1,23 +1,20 @@
-
-function startConverting()
-{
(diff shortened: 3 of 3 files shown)


Ritvyk:
@MaxReinisch can you please review it.
further advancements would be "Voice Command" that will be running in the background and when the user will say
>>Save page as "name" 
the page will be automatically archived.

sr6033:
Appending https:// automatically is not the right way to do it. Different pages can have different protocols. So, we cannot generalise that by our means.

Ritvyk:
@sr0633 I think we must discuss over the feature that enables the user for voice commands. , https:// or for http:// we could have a different solution.

MaxReinisch:
I shared this PR with one of my seniors, Eugene.  We both express concerns whether this feature would be useful in the extension, or if the extension is the right place for it.  He also had some suggestions regarding implementation:

> About implementation. I wouldn't recommend to merge as it is, and at least would request some changes:

> 1) it is implemented in incompatible way - there is more general API then `webkitSpeechRecognition` https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API/Using_the_Web_Speech_API

> 2) code style looks like java not javascript style `{` from new line, not spaces between operations and arguments and etc.

> 3) and error prone logic inside of `recognition.onresult` handler (edited) 

That being said, even if it isn't a great fit for this project, if you are interested in voice apps, we have a different project you might want to check out.  Our [Google assistant / Alexa app ](https://github.com/internetarchive/internet-archive-voice-apps) is still in development.  Maybe they'd like to add WBM features to the voice apps.

thisisayush:
I agree with @MaxReinish, the feature is good but the use case needs to be more broad or say defined. Doing what can be done with a click of a button, and even to do that you have to click a button? Just saying.

Ritvyk:
@thisisayush @MaxReinisch , I am thinking now to run this feature in the background , That would be good and user will not be in a need to click on the button.
I am thinking to add : 
>>hello! internet archive 
for switching the voice feature on like Alexa or Google
>>Then it will accept the commands like save now.
is it good ?

thisisayush:
What we can do instead is implement the google actions "Play Jizz", etc on the extension. But again that's a out of scope implementation in wayback machine's chrome extension.
But implementing voice recognition just for saving the page or doing things that can be done with a single click doesn't seem of much use.

Ritvyk:
@MaxReinisch I have edited the code style and implemented the speech API that you suggested me.

Ritvyk:
@thisisayush i totally agree with you but here we are not talking about only "Save Now " voice command but instead we can play more with this feature and adding more functionalities to it.
for an instance "Open Internet archive" and "Show My Profile"
and what is good about is that 
>>the blind people can easily have access to Wayback Machine because they use screen text to speech converters to read any webpage.

anishsarangi:
Thank you for this PR. But now we are moving in different direction regarding extension features(_This feature is not in the todo list till now_). We can close this PR. If you feel any other improvements need to be done, please make a PR.

PR #359, opened 2019-04-20
Redesigned tag cloud feature

tanweer919:
Redesigned the tag cloud page.
**Screenshot**
Peek 2019-04-20 20-14

PR #372, opened 2019-05-02
Revert "Redesigned tag cloud feature"

MaxReinisch:
Reverts internetarchive/wayback-machine-chrome#359


MaxReinisch:
Reverted because the tags did not scale with frequency

PR #373, opened 2019-05-03
Now words in the tagcloud scale according to their frequency

tanweer919:
Now words will scale according to their frequency of occurence in the tagcloud.
Screenshot:
New Peek 2019-05-04 01-14


tanweer919:
@MaxReinisch can u review this PR??

MaxReinisch:
I'm noticing that the cloud box is not centered on the page, and the left edge doesn't look flat.  I think it has to do with the size of the containing element.  Be sure to test this when the size of the window is changed.

tanweer919:
@MaxReinisch corrected the issue.

anishsarangi:
@tanweer919 Thank you for this PR. But now we are moving in different direction regarding this UI and feature. We can close this PR. If you feel any other improvements need to be done, please make a PR.

PR #374, opened 2019-07-05
fixed hyperlink support issue

Zerichen:
Previously the hyperlink within the Internet Archive logo is not working. Now it will create a new tab of Wayback Machine homepage.


MaxReinisch:
Looks like you implemented 2 solutions by both adding an anchor tag and a click event on the logo.  Only 1 should be needed, correct? I'd suggest using the anchor tag, as it will turn the cursor into a finger to show the user that it is clickable.  If you want it to open a tab, make [target=_blank](https://www.w3schools.com/tags/att_a_target.asp)

Zerichen:
I think I have removed the anchor tag in index.html. As I tried before, the anchor tag did not work very well. Nothing happens after I clicked the logo, so I just replaced it with a click event.

MaxReinisch:
Ah my mistake, I think I had gotten confused between trying it and looking at the code changes.  

In this PR, there is no indication that the icon is clickable (ie, the cursor doesn't change when hovering over the icon). Let's have the icon look clickable.  I noticed the same goes for the tv news results.  I'll make that change.

PR #376, opened 2019-07-05
socialSharing now open by tab

Zerichen:
Using the window to open the social sharing feature is not a good experience so I changed it into tab by default and can follow the "show context" setting.


MaxReinisch:
Since we decided we are going to make everything tab by default, your second commit is now depricated.

Zerichen:
Sure, let's just leave this PR for now. I will see if I could encounter more URL that needs to be excluded.

PR #377, opened 2019-07-05
improving error page

Zerichen:
Now the extension will catch the specific error code(404, 408, 410, 451, 500, 502, 503, 504, 509, 520, 521, 523, 524, 525, and 526) with a corresponding message displayed on the title bar. UI is also being changed a bit.


MaxReinisch:
How have you been testing this? I am trying to figure out a good way.

Zerichen:
I found some fake API from google, you can use them to configure the status code and route path. However, some of them only allow a certain amount of API requests for free.
Try these two: 
https://beeceptor.com/
https://www.mocky.io/

Another way is to simply build a backend using an express framework to send those HTTP status codes.

PR #380, opened 2019-07-18
Create Resource label 'R'

Zerichen:
In this PR, I have added the 'R' label standing for Resource: 
- Amazon Books
- Wikipedia Books
- Wikipedia Cited papers
- TV News

And now all of these features will be toggled by on checkbox.

### DIFF
--- a/scripts/amazon.js
+++ b/scripts/amazon.js
@@ -1,7 +1,0 @@
-// Get book on amazon page through
-// https://archive.org/services/context/amazonbooks?url=...
-function get_amazonbooks (url) {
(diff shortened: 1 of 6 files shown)


MaxReinisch:
Looks great, however I noticed 1 issue.  The wikipedia books feature isn't running:

___
Screenshot 2019-07-18 at 4 58 12 PM 

Notice how in the current version, there is an archive icon for quick access to an archive link.  
___
Screenshot 2019-07-18 at 5 00 40 PM

With this change, that icon goes away.  

Everything else looks great.  Bring those icons back in if you can and I'll happily merge.

PR #385, opened 2019-07-25
Context button now will be greyed out and have a tooltip if none of the context is selected

Zerichen:


MaxReinisch:
Still isn't going away, and it might have to do with the review I just made.

PR #414, opened 2019-09-08
Refactoring global_url related code

Zerichen:
Previously, in `popup.js`, we are using a variable called `global_url` to store the current tab URL, however, the way we acquire the URL is by calling the `chrome.tabs.query` method, which is an asynchronous function. Hence we might fail to acquire the tab URL sometimes.

Now every time we need the tab URL, we just simply calling this method and do the following operations asynchronously.


MaxReinisch:
Please include a more descriptive title & description.

MaxReinisch:
Thanks for making the name change.  I think it looks good.  Fix the exiting conflicts and I'll merge

PR #426, opened 2020-02-10
URL_error_handling_fixed 

### LINKED ISSUE #366: Improving Error Page Handling
Right now, the extension catches any error code except 200 to show a wayback machine popup to see the most recent version of the page. This needs to be improved, for e.g. checking if a saved copy is available before showing the screen and also checking it for specific error codes, checking for 401 Unauthorised makes no sense.

Also, the UI for the dialog also needs to be improved, and instead of redirecting to the page, a better way would be to inject a full screen popup so that user can still have the reload feature.

ZaidBJ:
Earlier annoying pop up used to show when the URL is  not available, now a prior ajax call is done, to let the user know beforehand about the "unavailability of URL".

This fixes #366

### DIFF
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -33,6 +33,7 @@ function rewriteUserAgentHeader(e) {
   return { requestHeaders: e.requestHeaders };
 }
 
+
 function URLopener(open_url, url, wmIsAvailable) {
   if (wmIsAvailable === true) {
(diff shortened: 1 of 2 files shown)


anishsarangi:
Thank you @ZaidBJ for your PR. We have already added this feature in a different way. I am closing this PR.

Issue #425, opened 2020-02-08
Clickable buttons doesn't seem to be good


ZaidBJ:
Clickable buttons doesn't seem to be good, some hover effect on the buttons will do the job,

rishipurwar1:
I want to work on this issue

SundeepChand:
Has it been assigned to someone? If not then I'd like to work on this.

PR #427, opened 2020-02-21
Better styling for buttons

ZaidBJ:
Hover effect is added on buttons , earlier  buttons do not have any effect on hovering , this leaded to poor experience for  users

improved #425

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -37,6 +37,7 @@ li{
 }
 button {
   display: block;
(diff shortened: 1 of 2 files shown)


anishsarangi:
Thank you for this PR. We are now planning for making the UI same for all browser extension, may be like the UI of safari wayback machine chrome extension. We can close this PR. If you feel any other improvements need to be done, please make a PR.

Issue #302, opened 2019-03-28
New feature: Dark mode


cheevu-RS:
I was thinking that it would be cool to add a dark themed mode throughout the extension using css's invert filter on styles.css. It can be added as an option in the settings.
I can put a PR for this if the mentors are fine with this addition.

MaxReinisch:
Adding a dark mode is a cool idea, but using the invert filter might not be the best way to do it.  Check out the archive.org homepage and maybe try to copy the color scheme of that.  Also maybe this feature would best live in css/common.css

cheevu-RS:
Sure, will do that. Thanks!

cheevu-RS:
@MaxReinisch You want me to add the color scheme of archive.org as the dark theme?

cheevu-RS:
@MaxReinisch The archive.org webiste only has a grey white theme.
![image](https://user-images.githubusercontent.com/33937406/55274162-5e11bb00-52fb-11e9-8946-702af7077d5c.png)
What i am suggesting is more like..
![image](https://user-images.githubusercontent.com/33937406/55274158-420e1980-52fb-11e9-8ec4-d29a9393775c.png)
 It is a bit crude, but i can fix that.

sr6033:
Dark mode is really cool and keeping it in sync with the Internet Archive's theme is a nice idea. Nice work @cheevu-RS 
I will suggest you edit the logo as per the theme and also to choose such a colour that the texts are more clear.

cheevu-RS:
Sure, thanks @sr6033!

ZaidBJ:
@MaxReinisch , can I take up this issue

anishsarangi:
This feature we are currently implementing!!

PR #429, opened 2020-02-23
Dark mode enabled

ZaidBJ:
this PR introduces dark mode that can be switch by toggle button, material design guidlines were followed for the same

this improves #302


anishsarangi:
We can close this PR. Reason is same as in #427. Thank you for your effort though :-)

PR #435, opened 2020-02-26
cursor should change to pointer when hovered over social media icons …

Jayaprakashkumar:
Fixed#428


anishsarangi:
Thank you for this PR. We are now planning for making the UI same for all browser extension, may be like the UI of safari wayback machine chrome extension. We can close this PR. If you feel any other improvements need to be done, please make a PR.

PR #444, opened 2020-03-26
Design Enhancements to the about page

blackviking991:
**This is the page looking before I made the changes**:
Screenshot from 2020-03-26 16-31-42
**And this is how it looks now:**
Screenshot from 2020-03-26 17-16-17

I will further update the documentation and will work on other issues.


cgorringe:
We went with a different design, and rewrote the About page. Thanks for the effort!

PR #445, opened 2020-03-27
Update index.css

deadpool1999:
Just changed the cursor to pointer.{on hover}


anishsarangi:
Thank you for this PR. This fix makes sense !! Please take care of the indentation also.

cgorringe:
Thanks for the fix. Since we're currently redoing the UI, this was fixed elsewhere.

PR #462, opened 2020-04-17
Fix Back Button not visible from settings>>context

### LINKED ISSUE #431: Back Button not visible from settings>>context
**Describe the bug**
There is no way to toggle back from the context sub-setting

**To Reproduce**
Steps to reproduce the behaviour:
1. Open the extension
2. Click on settings
3. Select 'Context' Sub-option

**Expected behavior**
The back button should be visible.

**Screenshots**
error1

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser chrome
 - Version 80.0.3987.116

tikhsuP:
Please review

### DIFF
--- a/css/settings.css
+++ b/css/settings.css
@@ -1,7 +1,7 @@
 #settingPage {
     height: 458px;
     max-height: 458px;
-    overflow: hidden;
+    overflow: auto;
 }
 
 #general {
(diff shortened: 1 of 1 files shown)


cgorringe:
Thanks for the fix, although we are currently redoing the UI so this will not be applied.

PR #463, opened 2020-04-19
New ui of context screen and handled api error in overview context screen

anishsarangi:
I have added the new/updated context screen UI in this PR.
Moreover I handled the API failure message in the wayback overview context


cgorringe:
Looks fine.

PR #464, opened 2020-04-19
Fix/context url bug

anishsarangi:
This PR fixes the bug due to which in the background.js console we are getting an error tabIdPromise is not defined


cgorringe:
Thanks for the fix.

PR #466, opened 2020-04-20
UI Overhaul to match Safari Extension

cgorringe:
This changes the UI for the Main view and Settings, mostly index.html .css, and script files popup.js and settings.js. 

The "Save Page Now" checkbox options (Outlinks & Screenshot) are displayed but will be implemented in a future PR. 

Dark Mode styles implemented in CSS auto apply when macOS preferences change, otherwise it's set to Light Mode as default (e.g. in Windows.)

There's a slight bug in the "Show Context" button being enabled when it shouldn't. There may be other minor bugs to look out for.

chrome-all


tikhsuP:
This is a better look.

PR #467, opened 2020-04-21
Created a new Style Guide

cgorringe:
Please take a look and let me know what you think.


cgorringe:
These are related: #338 #325

cgorringe:
- Changed guide to suggest omitting semicolons in JS code. 
- Added an ESLint section and modified rules so eslint can be used now.

PR #486, opened 2020-05-03
Wayback Count feature

cgorringe:
This will retrieve the count of snapshots of the current page from the Wayback Machine and display it as a badge on the toolbar, and inside the popup. Enable from Settings. This is similar to the Safari extension implementation.

Chrome_Wayback_Count


anishsarangi:
What will happen if the user has enabled to see "Relevant Archived Resources" and "Wayback Counts" and at amazon books page (e.g https://www.amazon.in/End-Days-Predictions-Prophecies-Thorndike/dp/1410407462). 
Which badge will be shown ? "R" or "Wayback Counts"

tikhsuP:
Initially, it shows "R" and when the "Wayback Count" gets fetched, it shows that.
You can see, when it shows "R", there is no count shown above the Search Bar button.

Before  Count Fetched       |  After Count Fetched
:-------------------------:|:-------------------------:
Screenshot (70)_LI  |  InkedCapturekkkkkk_L

anishsarangi:
@tikhsuP In the second screenshot where it the "Read this Book" button ?

tikhsuP:
@anishsarangi 
I am not sure what is wrong. Maybe, it isn't displayed when the Count is shown. 
Let's wait for @cgorringe  to reply.

scripts/background.js
```diff
@@ -342,7 +343,16 @@ chrome.runtime.onMessage.addListener(function (message, sender, sendResponse) {
...
+    // wayback count
+    chrome.storage.sync.get(['wm_count'], function (event) {
+      if (event.wm_count === true) {
+        updateWaybackCountBadge(tab.id, tab.url)
+      } else {
+        updateWaybackCountBadge(tab.id, null)
+      }
+    })
+    // auto save page
     chrome.storage.sync.get(['auto_archive'], function (event) {
```
anishsarangi, line 356: We can merge this and the above `chrome.storage.sync.get(['wm_count'],function(event))`  in a single line as` chrome.storage.sync.get(['wm_count', 'auto_archive'],function(event){}`  and do all the required checks there.

anishsarangi:
@tikhsuP That should not happen. Both are our extension's feature. It is happening due to the change of "Badge Text". You can check the function `borrow_books` in popup.js  to get more idea why it is happening. :-)

scripts/background.js
cgorringe: Thanks I'll make this change.

cgorringe:
It's true I haven't accounted yet for the conflict in the badge text changing. I'm going to try to see if I can replace the "R" and "S" badges with a toolbar icon change so that they can work with the Wayback Count badge at the same time.

The "Saved 0 times." in the screenshot above is actually supposed to say "This page was never archived." and not display a badge, so I'm glad that you posted that screenshot! Got that fixed now.

I'm not sure if the "Read This Book" button is supposed to hide in the 2nd screenshot. Good catch. Will take a look!

anishsarangi:
@cgorringe for the time , you can actually remove `result.includes('R') ` in line no- 315 of popup.js, later we can think of a solution to optimise this things.

cgorringe:
Much fixed. Got the toolbar icon to update properly.
You can test the Resources by going to Wikipedia, or an Amazon site like:
https://www.amazon.in/End-Days-Predictions-Prophecies-Thorndike/dp/1410407462

Here's a screenshot:
toolbar-icon-example_may11

cgorringe:
Will merge this now, even though there are a few bugs, we'll cover in a new Issue.

PR #496, opened 2020-05-15
WIP: Change host URL by checking the browser

anishsarangi:
In this PR , I have changed the host url by checking the browser. This is still under development.

### DIFF
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -14,7 +14,7 @@ var VERSION = manifest.version
 var globalStatusCode = ''
 let toolbarIconState = {}
 let tabIdPromise
-var WB_API_URL = 'https://archive.org/wayback/available'
+var WB_API_URL = hostURL+'wayback/available'
 var newshosts = [
   'www.apnews.com',
   'www.factcheck.org',
@@ -60,7 +60,7 @@ function savePageNow(page_url, silent = false, options = []) {
       setTimeout(() => {
         reject(new Error('timeout'))
       }, 30000)
-      fetch('https://web.archive.org/save/',
+      fetch(hostURL+'save/',
         {
           credentials: 'include',
           method: 'POST',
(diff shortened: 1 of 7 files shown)


cgorringe:
I wonder if `InstallTrigger` would work in the future? Would using the user-agent string be better, such as:
`var isFirefox = (navigator.userAgent.indexOf("Firefox") !== -1)`

anishsarangi:
Thank you @cgorringe . I agree, it would be better if I will user the `userAgent` string  instead of `InstallTrigger`

cgorringe:
I haven't tested this yet @anishsarangi, but if it's ready, take it out of Draft if you want me to test and merge it.

anishsarangi:
No, I am still working on it. Will let it know when it is ready for review ! Thank you !

anishsarangi:
@cgorringe  WE can merge this now so that we can test all the endpoints !

cgorringe:
OK, merging so we can test it on master!

scripts/background.js
```diff
@@ -14,7 +14,7 @@ var VERSION = manifest.version
 var globalStatusCode = ''
 let toolbarIconState = {}
 let tabIdPromise
-var WB_API_URL = 'https://archive.org/wayback/available'
+var WB_API_URL = hostURL+'wayback/available'
```
cgorringe, line 17: Already merged, but found one possible bug here, as this may be defined before hostURL is?

anishsarangi: @cgorringe Actually utils.js is loaded before background.js . You can check in manifest.json file lineNo-37

anishsarangi:
@cgorringe As there are a lot of APIs we are using in the extension, I am thinking of defining all the URL in the utils.js itself and use that const variable in other places directly. That will increase the code readability. What do you think of this ?

PR #498, opened 2020-05-17
Improve Responsiveness of Contexts Page

tikhsuP:
Contexts Page is more responsive now.


anishsarangi:
Nice Work!!
Thank you.

PR #499, opened 2020-05-17
Show related sites when data is there in Alexa Context Screen

anishsarangi:
This PR includes fix to a bug in alexa context screen. Now Show Related data will show when data is there.


tikhsuP:
Working fine : )

PR #500, opened 2020-05-18
Shortened Count Badge to work on Firefox

cgorringe:
Please test on Firefox on a variety of websites.
Try to see if you can test against multiples of 10, such as: 100, 1K, 10K, 100K, 1M.
The goal is for the text to be fully visible, without any of it cut off.


cgorringe:
Tested thanks to help from @tikhsuP

PR #501, opened 2020-05-18
Improve the content representation in Annotations Screen

tikhsuP:
This PR improves the content representation of the Annotations Context Page.

### Current Page

1. The text layout is distorted due to `word-break: break-all;`
The words are being broken at the right border of the box due to this property.
2. The "There are no annotations for the current URL." is at center now after removing some CSS classes.

New Annotations Page

### New Page
Removing `word-break` and some classes provide a fix.

Capture

### DIFF
--- a/css/context.css
+++ b/css/context.css
@@ -92,7 +92,6 @@ p {
     font-size: 20px;
     line-height: 1.42857143;
     color: #333;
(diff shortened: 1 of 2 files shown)


cgorringe:
I tested it and it looks good.

PR #503, opened 2020-05-19
Show time when Save Page is within last 24 hours

cgorringe:
This also localizes the date or time string shown when hovering over the Save Page Now button.


tikhsuP:
Works nice : )

anishsarangi:
Good to go ! Thank you @cgorringe

Issue #507, opened 2020-05-20
Review and Reduce API calls + Privacy setting


cgorringe:
This idea here is to make more efficient by reducing API calls, and increasing privacy.

- Analyze API calls that are being made in the background.
  - How often? Are they excessive? Are the same calls being made repeatedly?
  - Can you find a way to reduce or cache repeated API calls?

- Add a Setting to top of General Settings, called "Private Mode".
  - This would work similar to "Show All Contexts" in that when checked, other items that use APIs such as "Wayback Count", "Auto Save Page", "Enable Relevant Archived Resources", "Auto-update Contexts", and "404 Not Found" will all be unchecked.
  - If user checks one of those, "Private Mode" will auto-uncheck.
  - Ensure that when "Private Mode" is checked, that no background API calls are made, even when user hovers over buttons, unless user explicitly clicks on a button (only then is it OK to make an API call.)

Dependent on #506 done first.

cgorringe:
Here's an example API call that happens repeatedly.

An authentication check is done every time user taps toolbar icon and popup opens.

- POST to https://web.archive.org/save/ - with empty parameters.
  - fetched from auth_check() in background.js
  - which is called from checkAuthentication() in popup.js
  - which is called from last_save() in popup.js
  - which is called from checkExcluded() in popup.js
  - which is called from window.onloadFuncs[], every time popup displays.

- Could this be reworked so that the API call is only done once, then cached in a bool var?
  - May need to test for offline status?
  - Test every time browser closes and reopens, so don't store in persistent storage.

mitchell-frost:
I am new to open source. Can I work on this?

PR #508, opened 2020-05-20
Bug fix for SPN and auto-save

cgorringe:
- Save Page Now button works again.
- Auto-save works.
- Toolbar icon changes to red 'S' during save, then a checkmark when finished saving.

### DIFF
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -55,7 +55,7 @@ function URLopener(open_url, url, wmIsAvailable) {
 function savePageNow(tabId, page_url, silent = false, options = []) {
   if (isValidUrl(page_url) && isNotExcludedUrl(page_url)) {
     const data = new URLSearchParams()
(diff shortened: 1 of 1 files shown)


cgorringe:
Realized I need to do 2 more things:
- [ ] Test and Abort auto_save() if no Internet connection (i.e. offline).
- [ ] Ensure that auto-save is disabled if not logged in.

anishsarangi:
@cgorringe The save API will not work if you have not been logged in.

anishsarangi:
Good to go !! Working fine for me .

cgorringe:
Currently, if not logged in and when auto-save is enabled, it'll briefly show the 'S' icon as it calls the save API, then quickly go back to the default icon. It doesn't cause any issues so I'm going to leave it as-is for now. We can revisit checking first if already logged-in after completing #507.

PR #511, opened 2020-05-20
Show message when URL is in the list of excluded urls

### LINKED ISSUE #510: Showing user a message that the extension will not work for some particular URL
When the user is on an URL that is in the “excluded-list”, he can be confused why some buttons are not working.
E.g-  If the user is on https://chrome.google.com/webstore/detail/wayback-machine/fpnmgdkabkmnadcjpehmlllkndpkmiak/reviews?hl=en
then he clicks on the extension and if he click on “facebook” sharing button, it will not work. But when the user will hover on “Save Page Now” button, then he will be seeing the message that “The URL is not supported”.
How about adding a “text” above the “input-box” that “The URL is not supported” ? OR may be at any other place so that the user will understand he can’t take any actions in that URL ?

anishsarangi:
In this PR, I have added a message "URL not supported" that will be shown above the input-box (stating that most the features of our extension will not work) when the URL is part of excluded URLs.
Fixes #510

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -57,7 +57,7 @@ hr {
     color: #222;
 }
 
-#wayback-count-label {
+#wayback-count-label, #url-not-supported-message {
     font-size: 9pt;
     color: #222;
 }
@@ -337,7 +337,7 @@ a.btn-social, .btn-social {
         color: #ddd;
     }
 
-    #wayback-count-label {
+    #wayback-count-label,#url-not-supported-message {
         color: #eee;
     }
 
--- a/index.html
+++ b/index.html
@@ -23,7 +23,7 @@
       <div>
         <div class="text-center">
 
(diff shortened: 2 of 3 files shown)


cgorringe:
Surprised that it only took 2 lines, but it's effective!
I have 2 suggestions:
- Start using dashes (-) instead of underscores (_) for ids. (see style guide).
- Add CSS style to index.css so that it matches font-size and color of `wayback-count-label`. Or just add it to that one. Don't forget the Dark Mode version.

anishsarangi:
Thank @cgorringe . Working on it !

anishsarangi:
@cgorringe I have made the changes. Please have a look over it !

cgorringe:
I'm having trouble pulling the latest changes from Github. I'll merge this to master since it looks OK.

PR #513, opened 2020-05-21
Enabled tagcloud feature after the API fix

anishsarangi:
@cgorringe I have enabled again "Tagcloud" feature in the context. Thank you !


cgorringe:
Haven't tested, but looks fine. Merging.

PR #514, opened 2020-05-22
Tweaked styles for Tag Cloud context page

cgorringe:
Just did some CSS tweaks to change background color, add a rounded border, etc.


anishsarangi:
Good to go !!

Issue #512, opened 2020-05-21
Auto-update Context should remain on same context


cgorringe:
**Describe the bug**

When auto-updating contexts, when you switch between different websites, the Context page does correctly reload. However, the highlighted context reverts to Alexa Internet or whatever's the first tab.

**To Reproduce**
1. Check "Auto-update Contexts on New Pages".
2. Check "Show All Contexts".
3. (Optionally) Show Features in Window.
4. Open a website, then open in a new tab or window, another website.

**Expected behavior**
I'd like to see that in the Context window, remain on the same context. For example, if I were previously on "Annotations" I'd like to see it remain there when auto-updating.

**Desktop**
 - OS: Mac 10.14.6 Mojave
 - Chrome Browser
 - Version 81.0.4044.138 (Official Build) (64-bit)

PR #515, opened 2020-05-22
Reopen same context tab on reload

tikhsuP:
This PR solves issue #512.

### DIFF
--- a/scripts/context.js
+++ b/scripts/context.js
@@ -42,6 +42,7 @@ function openContextFeature(evt, feature) {
   // Show the current tab, and add an "active" class to the button that opened the tab
   $(feature).show()
   evt.currentTarget.className += ' active'
+  return(feature)
 }
 
 function singlePageView() {
@@ -69,9 +70,15 @@ function singlePageView() {
       } else {
         contexts_dic[feature]()
         $(featureTabId).click(function(event) {
-          openContextFeature(event, featureId)
+          let featureTabId = openContextFeature(event, featureId) + '_tab'
+          chrome.storage.sync.set({featureTabId: featureTabId}, function() {
+            console.log('Value is set to ' + featureTabId);
+          })
         })
         if (!first_feature) {
+          chrome.storage.sync.get(['featureTabId'], function(result) {
+            console.log('Value received from storage' + result.featureTabId);
+          })
           first_feature = featureTabId
         }
       }
--- a/scripts/context.js
+++ b/scripts/context.js
@@ -42,7 +42,7 @@ function openContextFeature(evt, feature) {
   // Show the current tab, and add an "active" class to the button that opened the tab
   $(feature).show()
   evt.currentTarget.className += ' active'
-  return(feature)
+  return feature
 }
 
 function singlePageView() {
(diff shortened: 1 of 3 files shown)


scripts/context.js
```diff
@@ -69,9 +70,15 @@ function singlePageView() {
       } else {
         contexts_dic[feature]()
         $(featureTabId).click(function(event) {
-          openContextFeature(event, featureId)
+          let featureTabId = openContextFeature(event, featureId) + '_tab'
```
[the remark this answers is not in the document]
tikhsuP: I am making it to return the 'feature' parameter, which is further being stored using chrome storage.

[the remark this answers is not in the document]
anishsarangi: Ah... !! I missed the above top line in `changes`.

scripts/context.js
```diff
@@ -42,6 +42,7 @@ function openContextFeature(evt, feature) {
   // Show the current tab, and add an "active" class to the button that opened the tab
   $(feature).show()
   evt.currentTarget.className += ' active'
+  return(feature)
```
anishsarangi, line 45: remove brackets

tikhsuP: okay, thanks.

anishsarangi:
In chrome.tabs.onActivated `chrome.tabs.update(id, { url: chrome.runtime.getURL('context.html') + '?url=' + tab.url })` pass the featureID as a URL parameter. AND in the context.js you can get the URL parameter as make that clicked automatically.

tikhsuP:
@anishsarangi I tried a different solution. I was already working on that. Please have a look. 
Thanks :)

scripts/context.js
```diff
@@ -69,16 +69,21 @@ function singlePageView() {
...
         contexts_dic[feature]()
         $(featureTabId).click(function(event) {
-          openContextFeature(event, featureId)
+          let openedFeature = openContextFeature(event, featureId) + '_tab'
+          chrome.storage.sync.set({openedFeature: openedFeature}, function() {
+          })
+        })
+        //Set the default feature to be opened
+        chrome.storage.sync.get(['openedFeature'], function(result) {
+          openedFeature = result.openedFeature
```
anishsarangi, line 78: Can we rename this variable ? And please initialise the variable properly. Thank you

anishsarangi, line 78: Can we rename this variable ? Then same named variable is used in line no-72 And please initialise the variable properly. Thank you

anishsarangi:
@tikhsuP Add 'moz-extension://' in the list of excluded URLs in `utils.js` to make it work properly in firefox.

scripts/context.js
tikhsuP: Apologies. Correcting it. Thanks

anishsarangi:
Looks fine :-)

cgorringe:
I did find the rare case where if you, for example, Show All Contexts (in Settings), Show Contexts and select the Annotations tab, then clear Annotations in Settings. Next time you Show Contexts it'll show an empty page with no highlighted tabs. It's not that big of a deal, since the user can just select one of the tabs and it'll work fine.

Looks good, so I'll merge it.

PR #516, opened 2020-05-22
Fix loader bug in tagcloud context screen

anishsarangi:
NO loader was showing up when the API call was taking some time. I have fixed that

### DIFF
--- a/scripts/tagcloud.js
+++ b/scripts/tagcloud.js
@@ -24,8 +24,8 @@ function get_tags (url) {
   fetch(new_url)
     .then(response => response.json())
     .then(function (data) {
-      $('#container-wordcloud').show()
       $('#loader_tagcloud').hide()
+      $('#container-wordcloud').show()
       if (!data.error && data.length > 0) {
         for (let i = 0; i < data.length; i++) {
           var b = {}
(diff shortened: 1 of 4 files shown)


scripts/tagcloud.js
```diff
@@ -20,12 +20,13 @@ function get_tags (url) {
   var not_display3 = not_display4 + ' extension'
   var dontarray = ['view page', 'open', 'read more', not_display1, not_display2, not_display3, not_display4]
   var new_url = 'https://archive.org/services/context/tagcloud?url=' + toBeUsedAsURL
-  $('#loader_tagcloud').show()
+  $('#container-wordcloud').hide()
   fetch(new_url)
     .then(response => response.json())
     .then(function (data) {
+      $('#container-wordcloud').show()
```
cgorringe, line 27: the .show() on line 27 should be after the if on line 29.

anishsarangi: YES, correct. I am changing it sorry !1

anishsarangi:
I have made changes @cgorringe . Please review !

cgorringe:
Goes **after** the **if**:
`if (!data.error && data.length > 0) {`

anishsarangi:
@cgorringe Please check the else condition. I have already handled that case.

cgorringe:
Looks good.

PR #518, opened 2020-05-24
Add 'Enable 404 popup' in settings as a option

### LINKED ISSUE #506: "Enable 404 Not Found" Setting
I'd like to see an "Enable 404 Not Found" item added to General Settings.
- When set, turns on Wayback Availability check when visiting a 404 page, and displays HTML popup. This is what currently happens.
- When not set, turn off the check (don't call API), and don't display popup.
- This should be set default ON when user runs extension for first time.

Here's a 404 link to test against:
http://purl.oclc.org/docs/inet96.html

anishsarangi:
In this PR I have added 'Enable 404 popup' in "General Settings". By default it is "checked" , the users can turn off as per their needs.
Fixes #506 

Here's a 404 link to test against:
http://purl.oclc.org/docs/inet96.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -185,6 +185,12 @@
             <span class="toggle">Enable Relevant Archived Resources</span>
           </div>
         </label>
+        <label class="setting-switch" for="not-found-popup">
+          <input type="checkbox" id="not-found-popup">
+          <div class="setting-text-box">
+            <span class="toggle">Enable Popup for 404 Pages</span>
+          </div>
+        </label>
         <label class="setting-switch" for="auto-update-context">
           <input type="checkbox" id="auto-update-context">
           <div class="setting-text-box">
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -14,7 +14,7 @@ var VERSION = manifest.version
 var globalStatusCode = ''
 let toolbarIconState = {}
 let tabIdPromise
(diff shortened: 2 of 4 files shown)


cgorringe:
I might tweak the wording later, but overall this works fine! So I'll merge it.

tikhsuP:
I made the following observations:

1. This feature works even without clicking on the "Accept and Enable" button in the Welcome page. This should not happen.

2. The feature works well on Chrome, Firefox (Basic) but not on Firefox Developer Edition. 
When I paste the link in the Address bar, and click Enter, it doesn't shows the popup. Whereas when I click reload or when I am redirected to [this](http://purl.oclc.org/docs/inet96.html) page, popup appears.
The feature doesn't works only when I paste the URL directly.
This seem to be an issue with Firefox Developer Edition itself, and not in our Extension. But, please have a look.

3. We can add  `cursor: pointer;`  in  `#close { }`  in  `archive.css`.

cgorringe:
1. It's the default to be on, so by design, but we can see about turning it off before that button is pressed.
2. OK, I hadn't tested it on firefox.
3. sounds good

anishsarangi:
@tikhsuP Thank you for the observations. I am working on 1 and 3 . But I am not sure about 2, I will check that. Thank you !

PR #524, opened 2020-05-25
fixed error-404-popup bug

anishsarangi:
This PR fixes the following issues

1. This feature works even without clicking on the "Accept and Enable" button in the Welcome 
     page. This should not happen

2. We can add `cursor: pointer` in #close { } in archive.css.

In developer edition of firefox I am testing this, I got to know why it is happening, but still thinking of a solution.

### DIFF
--- a/css/archive.css
+++ b/css/archive.css
@@ -34,6 +34,10 @@
     right: 10px;
 }
 
+#close {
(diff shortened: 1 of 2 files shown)


cgorringe:
Looks fine.

PR #527, opened 2020-05-26
Add Link Sharing and Email Sharing

tikhsuP:
This PR adds the following to the share options:

 1. A link sharing option which, when clicked, copies the link of the archived page to the user's clipboard.
 2. A share through mail option.


anishsarangi:
For email sharing, I could only see one option that is using `mailto: `, and you can try binding dynamically binding that through popup.js when the page loads.!

scripts/popup.js
```diff
@@ -158,14 +165,31 @@ function social_share(eventObj) {
...
+          toastMsg[0].setAttribute('id', 'toast-success')
+          toastMsg[0].innerText = 'Page link copied.'
+          showToast(toastMsg[0])
+        }, function() {
+          toastMsg[0].setAttribute('id', 'toast-fail')
+          toastMsg[0].innerText = 'Error occurred. Try again.'
+          showToast(toastMsg[0])
+        })
+      } else if (id.includes('email')) {
+        $(document).ready(function() {
```
anishsarangi, line 180: Why didi you add `$(document).ready()`? It will still work if you will not add that!!

scripts/popup.js
```diff
@@ -498,6 +522,8 @@ $('#first_capture').click(first_capture)
 $('#fb_share').click(social_share)
 $('#twit_share').click(social_share)
 $('#linkedin_share').click(social_share)
+$('#copy-link').click(social_share)
+$('#email-share').click(social_share)
```
anishsarangi, line 526: Add the href attribute directly inside `window.onload` function !

tikhsuP: Thanks.

scripts/popup.js
tikhsuP: This function checks if the document is completely loaded or not. And performs the action only if it is needed. 

YES... there's no need to use this here. 
Thanks :)

scripts/popup.js
tikhsuP: I made the required changes. Is there a better way to do this?

scripts/popup.js
```diff
@@ -484,7 +504,23 @@ chrome.runtime.onMessage.addListener(
...
+  var overview_url = 'https://web.archive.org/web/*/'
+  chrome.tabs.query({ active: true, currentWindow: true }, function (tabs) {
+    let url = tabs[0].url
+    if (isNotExcludedUrl(url)) { // Prevents sharing some unnecessary page
+      if (url.includes('web.archive.org')) {
+        sharing_url = url // If the user is already at a playback page,share that URL
+      } else {
+        sharing_url = overview_url + get_clean_url(url) // When not on a playback page,share the overview version of that URL
+      }
+      $('#email-share').attr('href', `mailto:?subject=Hey, please check out this amazing website.&body=${sharing_url}`)
```
anishsarangi, line 518: You can change the subject here like "Hey, Please checkout all archives of this website available in web.archive.org"

tikhsuP: Sure, and, what about the "Toast" messages? Please tell me if you want any changes there :)

anishsarangi: Looks fine to me !

cgorringe:
Link Sharing got implemented in PR #683. We'll skip email sharing and close this now. Thanks!

PR #534, opened 2020-05-30
Code added to work the extension in firefox

anishsarangi:
In this PR I have added some code to make this extension work in firefox !


tikhsuP:
Working fine!!

cgorringe:
OK, that was what I commented out previously, and thanks for verifying that part is for Firefox. I'm fixing this in PR #532

PR #535, opened 2020-06-03
Hides URL address bar for contexts shown in a new window

cgorringe:
As suggested by Mark, this hides the address bar for Show Context in a new window. Doesn't (can't) hide if opened in a new tab.

### DIFF
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -262,9 +262,9 @@ function opener(url, option, callback) {
       if (callback) { callback(tab.id) }
     })
   } else {
-    let width = window.screen.availWidth
-    let height = window.screen.availHeight
-    chrome.windows.create({ url: url, width: width / 2, height, top: 0, left: width / 2, focused: true }, function (window) {
+    let width = Math.floor(window.screen.availWidth * 0.75)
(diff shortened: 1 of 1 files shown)


tikhsuP:
Works well in Chrome.!!
Please have a look at #536

scripts/utils.js
```diff
@@ -262,9 +262,9 @@ function opener(url, option, callback) {
       if (callback) { callback(tab.id) }
     })
   } else {
-    let width = window.screen.availWidth
-    let height = window.screen.availHeight
-    chrome.windows.create({ url: url, width: width / 2, height, top: 0, left: width / 2, focused: true }, function (window) {
+    let width = Math.floor(window.screen.availWidth * 0.75)
+    let height = Math.floor(window.screen.availHeight * 0.90)
+    chrome.windows.create({ url: url, width: width, height: height, top: 0, left: 0, focused: true, type:'popup'}, function (window) {
```
anishsarangi, line 267: Please remove `focused:true` as it is not supported in firefox as well as we don't need that for chrome also. It will fix the issue #536 . Thank you !

PR #538, opened 2020-06-04
Added extra urls to excluded list based on firefox browser

anishsarangi:
In this PR I have added some extra urls based on firefox browser!

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -20,12 +20,12 @@
       "64": "images/toolbar/toolbar-icon-archive64.png"
     }
   },
-//  "applications": {
-//    "gecko": {
-//      "id": "wayback@web.archive.org",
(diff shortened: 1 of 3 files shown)


cgorringe:
- How about exclude anything that starts with `about:` which would replace 3 cases on that list?
- `chrome://` already covers `chrome://newtab` so you can remove the newtab one.
- There are other IP ranges we can include. Let's add `192.168.` since that is reserved for local networks. Another one is `10.`

tikhsuP:
Yes, let's exclude anything that starts with `about:` 
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/The_about_protocol

cgorringe:
Looks good.

PR #539, opened 2020-06-05
Set Tabs to be Opened in Context Page

tikhsuP:
This PR provides the following fixes:

1. Open the first tab if the user is accessing the Context Page for the first time.
2. Open the previously opened tab when the user switches between webpages with **"Auto-update Contexts"** option enabled.
3. Open the first tab if the previously opened tab is not selected in the Contexts menu.
4. Open the previously opened tab if it's option is selected in the Contexts menu.

### DIFF
--- a/scripts/context.js
+++ b/scripts/context.js
@@ -56,7 +56,7 @@ function singlePageView() {
   // Check settings for features
   let features = ['alexa', 'domaintools', 'wbmsummary', 'annotations', 'tagcloud']
   chrome.storage.sync.get(features, function (event) {
-    let firstFeature = null
+    let clickFeature = null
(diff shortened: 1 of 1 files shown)


scripts/context.js
```diff
@@ -72,18 +73,29 @@ function singlePageView() {
           chrome.storage.sync.set({selectedFeature: selectedFeature}, function() {
           })
         })
-        //Set the default feature to be opened
         chrome.storage.sync.get(['selectedFeature'], function(result) {
```
anishsarangi, line 76: Please take this out of this for loop and declare a `var openedFeature` there !
Thanks

PR #542, opened 2020-06-09
Increase z-index in archive popup

anishsarangi:
In this PR I have increased the z-index of the 404-popup!

### DIFF
--- a/css/archive.css
+++ b/css/archive.css
@@ -1,6 +1,6 @@
 #error {
     position: fixed;
-    z-index: 100;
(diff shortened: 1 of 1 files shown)


tikhsuP:
Thanks for this upgrade.
Should we shift it a little lower too? Just to 'not hide' the content of the nav-bar.

Capture

cgorringe:
Ironically, Medium.com also uses a z-index of 500.
If you really want to be above everything else, use: 
`z-index: 2147483647`

PR #550, opened 2020-06-11
Fix 'Show Contexts' button bugs

tikhsuP:
This PR provides the following fixes:
1. Clicking 'Enable in Settings' opens Context settings.
2. Dimming the 'Show Contexts' button when disabled.

### DIFF
--- a/scripts/popup.js
+++ b/scripts/popup.js
@@ -393,10 +393,27 @@ function show_wikibooks() {
 function noContextTip() {
   chrome.storage.sync.get(['alexa', 'domaintools', 'tweets', 'wbmsummary', 'annotations', 'tagcloud'], function (event) {
     // If none of the context is selected, grey out the button and adding tip when the user hovers
-    for (const context in event) { if (event[context]) { return $('#context-screen').click(show_all_screens) } }
-    if (!$('#ctxbox').hasClass('flip-inside')) { $('#ctxbox').addClass('flip-inside') }
+    for (const context in event) { 
+      if (event[context]) { 
+        $('#contextBtn').removeAttr('disabled')
+        return $('#contextBtn').click(show_all_screens) 
+      }
+    }
+    if (!$('#ctxbox').hasClass('flip-inside')) {
+      $('#ctxbox').addClass('flip-inside')
+      $('#contextBtn').attr('disabled', true)
+    }
     /* $('#context-screen').css({ opacity: 0.5 }) */
   })
+}
+
+function openContextMenu () {
+  $('#popup-page').hide()
+  $('#setting-page').show()
+  $('#general-panel').hide()
+  $('#context-panel').show()
+  if (!$('#context-btn').hasClass('selected')) { $('#context-btn').addClass('selected') }
+  if ($('#general-btn').hasClass('selected')) { $('#general-btn').removeClass('selected') }
 }
 
 function checkExcluded() {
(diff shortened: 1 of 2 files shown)


cgorringe:
Doesn't yet dim the button when the popup opens, but does work when tapping Back from the Settings, then tapping on the dimmed 'Show Contexts' button. (still work in progress)

tikhsuP:
Instead of calling `chrome.tabs.query({ active: true, currentWindow: true }, function (tabs) { }` twice on lines 121 and 128, should I call it before the `if (noneSelected()) { }` statement?

scripts/popup.js
```diff
@@ -393,23 +393,41 @@ function show_wikibooks() {
 function noContextTip() {
   chrome.storage.sync.get(['alexa', 'domaintools', 'tweets', 'wbmsummary', 'annotations', 'tagcloud'], function (event) {
     // If none of the context is selected, grey out the button and adding tip when the user hovers
-    for (const context in event) { if (event[context]) { return $('#context-screen').click(show_all_screens) } }
-    if (!$('#ctxbox').hasClass('flip-inside')) { $('#ctxbox').addClass('flip-inside') }
+    for (const context in event) { 
+      if (event[context]) { 
+        $('#contextBtn').removeAttr('disabled')
```
anishsarangi, line 398: Why this is required ?

scripts/popup.js
```diff
@@ -393,23 +393,41 @@ function show_wikibooks() {
...
+        return $('#contextBtn').click(show_all_screens) 
+      }
+    }
+    if (!$('#ctxbox').hasClass('flip-inside')) {
+      $('#ctxbox').addClass('flip-inside')
+      $('#contextBtn').attr('disabled', true)
```
anishsarangi, line 404: Is this line is required ? Please tell me a situation where it is useful

cgorringe:
I haven't studied the code too closely, but I tried it out, and it looks like it all works!

You can delete the 2 lines of code and the 'TODO' that were previously commented out. (which I may have done in the past.)

anishsarangi:
It is working fine @cgorringe . I have tested it. :-)

scripts/popup.js
tikhsuP: This line is required if the user initially has no contexts selected, visits a non excluded URL and clicks on the extension icon. This line dims the 'Show Contexts' button in this case, same as the Back button. The dimmed button tells the user that no context is selected and the flip tells to select contexts in the settings.

scripts/popup.js
tikhsuP: This line is required if you initially have any context option checked in the context menu, go to an excluded URL (eg. newtab), and click on the extension icon. This line darkens the color of the 'Show Contexts' button if the user has some contexts selected and then visits on an excluded URL (Back button does the same). The dark 'Show Contexts' button tells the user that contexts are selected in the Context menu, and the flip tells the user that this is an excluded URL.

anishsarangi: I was not aware of, when to darkens the color of the 'Show Context' button. Anyway I got an overall idea. Thank you !

scripts/popup.js
anishsarangi: This make sense actually. Thank you for explaining !

tikhsuP: Thanks :-)

scripts/popup.js
tikhsuP: Thank you!! :-)

PR #555, opened 2020-06-13
Show error message on empty Context screen

tikhsuP:
This PR shows an error message to the user on the 'Context' page if there are no contexts selected.

### DIFF
--- a/context.html
+++ b/context.html
@@ -20,7 +20,7 @@
 </head>
 <body>
     <!-- Error message -->
-    <div class="container" id="error-message">
+    <div class="container display-none" id="error-message">
         <div>
             <div class="col-sm-12 text-center">
                 <h2>Error</h2>
--- a/context.html
+++ b/context.html
@@ -27,7 +27,7 @@
             </div>
             <div class="col-sm-6">
                 <div class="pre">
-                    <div class="color_code text-center"><h4>No Context Selected</h4></div>
+                    <div class="text-center"><h4 class="color_code">No Context Selected</h4></div>
                     <div class="error">
                         <span>Looks like you didn't make any selection in the <b>Contexts</b> menu. Please follow the steps given below:</span>
                         <ul>
--- a/context.html
+++ b/context.html
@@ -20,7 +20,7 @@
 </head>
 <body>
     <!-- Error message -->
-    <div class="container display-none" id="error-message">
+    <div class="container" id="error-message">
         <div>
             <div class="col-sm-12 text-center">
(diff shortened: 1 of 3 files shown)


scripts/context.js
```diff
@@ -59,6 +59,7 @@ function singlePageView() {
     chrome.storage.sync.get(['selectedFeature'], function(result) {
       var openedFeature = result.selectedFeature
       let clickFeature = null
+      let countFeature = 0
```
anishsarangi, line 62: Why don't you directly assign it to `features.length ` ?

scripts/context.js
```diff
@@ -94,6 +96,10 @@ function singlePageView() {
           }
         }
       }
+      //Hide error message if a context is selected
+      if (countFeature > 0) {
```
anishsarangi, line 100: Please hide it by default, and show it if the count is <= 0 . Thank you

scripts/context.js
tikhsuP: `let features = ['alexa', 'domaintools', 'wbmsummary', 'annotations', 'tagcloud']` this is `features` and it will always return `5` as its length.

scripts/context.js
tikhsuP: I would like to learn the difference between these two cases. Please explain. Thanks :)

anishsarangi: Whenever i reload the context screen i could see the not context selected popup for a sec. This will solve that issue

scripts/context.js
anishsarangi: OK, got it. Thank you

context.html
```diff
@@ -19,6 +19,29 @@
     <title>Show Contexts</title>
 </head>
 <body>
+    <!-- Error message -->
+    <div class="container display-none" id="error-message">
```
anishsarangi, line 23: Please give it a proper class name or add style to id `error-message`

scripts/context.js
```diff
@@ -94,6 +96,10 @@ function singlePageView() {
           }
         }
       }
+      //Show error message if no context is selected
+      if (countFeature <= 0) {
+        $('#error-message').removeClass('display-none')
```
anishsarangi, line 101: You can use show() and hide() method of jquery

anishsarangi: I mean add display none by default and you can use `hide` or `show` jquery methods instead of `addClass` and `removeClass`. I know using both the methods will result the same thing that was happening before. Thank you

PR #562, opened 2020-06-18
Handle error and show proper message if error happens while doing SPN

anishsarangi:
In this PR I have handled error while doing SPN and showing proper message for that in the UI.
Thank you !

### DIFF
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -140,6 +140,8 @@ async function validate_spn(tabId, job_id, silent = false) {
         status = data.status
         vdata = data
       })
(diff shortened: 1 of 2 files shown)


cgorringe:
I pushed a fix. I can add a toolbar icon in a future PR, after my other PR gets merged to master.

anishsarangi:
@cgorringe I couldn't test it properly as I could not find a situation where the SPN is failing, so was not aware of how long could be the error message. Your fix is great for this and covers all the cases. Thank you !
Toolbar Icon should also be added.

PR #571, opened 2020-06-22
Resource list and status shown in a new tab

anishsarangi:


cgorringe:
Cool to see this working!

Here's my review:
- Nitpick on CSS id's could use dashes (-) instead of underscores (_).

- But more importantly, you're calling the status API over again in `show_resource_data()`, but it's being fetched simultaneously in `background.js`. How about we take the data from the `validate_spn()` function to avoid the extra API calls? One idea is to call `chrome.runtime.sendMessage()` and send the `data` on line 147 of `background.js`, then retrieve the message in `resource_list.js` which can then output the URL list. This should also reduce the amount of code.

webextension/scripts/settings.js
```diff
@@ -144,11 +164,12 @@ function switchTabWindow() { $('input[type="radio"]').not(':checked').prop('chec
 
 function addDocs () {
```
cgorringe, line 165: Merge incorrect: Revert to prior on most lines. Only add the 'show-resource-list' line. I'd like to reword it to: 'Display a list of resources during Save Page Now.'

webextension/scripts/utils.js
```diff
@@ -329,9 +329,12 @@ function initDefaultOptions () {
     showall: true,
     alexa: true,
     domaintools: false,
-    wbmsummary: true,
-    annotations: true,
-    tagcloud: true
+    wbmsummary: false,
```
cgorringe, line 332: Revert to prior, only add `show_resource_list` line.

webextension/scripts/background.js
```diff
@@ -70,6 +70,12 @@ function savePageNow(tabId, page_url, silent = false, options = []) {
       .then((res) => {
         if (!silent) {
           notify('Saving ' + page_url)
+          chrome.storage.sync.get(['show_resource_list'], function(result) {
```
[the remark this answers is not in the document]
tikhsuP: Yes, because of this, I cannot see the resources list even when the option is checked. Changing it to `chrome.storage.local` will solve this :-)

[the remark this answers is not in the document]
anishsarangi: I have fixed this issue. !!

webextension/scripts/settings.js
```diff
@@ -19,6 +19,24 @@ switchSetting()
 addDocs()
 
 function initializeSettings () {
```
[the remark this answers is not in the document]
anishsarangi: I have reverted the changes. Thank you !!

webextension/scripts/settings.js
anishsarangi: I have reverted the changes !!

anishsarangi:
I have added the message when the tab is reloaded. @cgorringe Please check !
Thank you.

anishsarangi:
@tikhsuP You can go ahead for adding styles if you want ! Thank you .

cgorringe:
I'm a little confused on the use of Set() and comparing `new_resource_length > old_resource_length` so going to look again tomorrow after I get some sleep...

cgorringe:
Good reduction in code, looks better!
1. I noticed that you didn't handle the message for `resource_list_show_error`.
2.`$('.error').hide()` on line 15 should be moved to inside the `if` as there are different kinds of messages that may be received that are unrelated to resource lists.
3. I'd shrink the size of the URL, since they can get long, so maybe smaller than h2. Could be on a line following "Saving Page" title.

tikhsuP:
I made some changes, will push more changes and make it similar to `context.html `once that is ready.

anishsarangi:
@cgorringe  I have used Set to store the urls that are coming as the resources . 
Before I was checking both the previous_length and new_length to know if any additions are there in the Set or not. If Yes, then I am appending to the div . If Not, then I was not appending dynamically !

cgorringe:
Also I think this looks nicer as a popup Window and not a new tab, so maybe you can force it to window despite the window/tab setting?

cgorringe:
This was shown when status = `error`

Screen Shot 2020-07-04 at 10 25 28 PM

cgorringe:
Could you also check `message.data.message` (`message` key in json) and output the message just below "Success" if it exists?

cgorringe:
We should have something animated during "pending" status. Like a spinner animated GIF.

cgorringe:
Instead of taking `status` from the API, uppercasing the first letter and outputting that, use a dictionary of status keys with values of what you want to output. This will work out better later on when we consider translating to other languages.

webextension/scripts/resource_list.js
```diff
@@ -16,14 +21,18 @@ function show_resource_data(url_name) {
...
+          $('#message').text(vdata.message)
+        }
         if ('resources' in vdata) {
           new_resource_length = vdata.resources.length
           vdata.resources.forEach((element) => {
             resource_list_data.add(element)
           })
         }
         if(new_resource_length>old_resource_length){
+          $('.loader').hide()
```
tikhsuP, line 35: The `status_list` and `#message` look really good to me.
Also, I would suggest hiding the loader inside `if (status === 'success')` instead of here.

anishsarangi: No, we also get the list of resource data sometimes when the status is pending !! So We have to show the loader here !

tikhsuP: Yes, I see that. I suggested it because the download is still not completed and removing the loader makes it feel as if it is completed. It is more like a choice, we can remove or keep the loader for this case. However, not hiding the loader will look like this, which may show the 'Processing'/pending/downloading state more clearly :-)

Screenshot (117)

anishsarangi: Oh, I got your point. Thank you ! I will add this in the next commit !

webextension/scripts/resource_list.js
```diff
@@ -0,0 +1,54 @@
...
+              $('<p>').append(item)
+            )
+          }
+          old_resource_length = new_resource_length
+        }
+        if (status === 'success') {
+          $('#counter-container').show()
+          new_resource_length = vdata.resources.length
+          $('#spn-elements-counter').text(new_resource_length)
+        } else if (!status || (status === 'error')) {
```
tikhsuP, line 47: Should we remove `!status` from this `if` statement?
Because, few seconds after a successful save, status becomes `undefined` which results in this. Is this expected?

Screenshot (118)

cgorringe:
- The list of resources is disappearing a short while after success.
- The `message` content may span multiple lines, in which case the words split mid-word due to `word-break: break-all` being applied to all `<p>`. Maybe you could apply word break only to those `<p>` inside the container?

My suggested status messages:

> const status_list = {
>   pending: 'Processing...',
>   success: 'Save Succeeded.',
>   error: 'An Error Occurred. Please Try Again.'
> }

anishsarangi:
@tikhsuP Please test it now. Thank you !

cgorringe:
I made some tweaks. Try resizing the window while / after the list of URLs display.

The **ONE** thing that's bugging me is now the spinning loader sometimes deforms into a squished rotating oval and I don't know why!

cgorringe:
The loader is already a fixed width & height in pixels, which should be fine (you would think!) but for some reason it changes. Hmm maybe if I add `!important` to them in the CSS... Nope. But I'll figure this out!

PR #572, opened 2020-06-22
Open Feedback URL Depending on Browser

tikhsuP:
This PR provides the following fix:
The **Feedback** button in the extension and the **Write a Review** option in the **About** page now open the appropriate feedback pages depending on the browser (Chrome/Firefox).

### DIFF
--- a/scripts/about.js
+++ b/scripts/about.js
@@ -1,6 +1,9 @@
-$(document).ready(function () {
-  const VERSION = chrome.runtime.getManifest().version
-  const DATE = new Date().getFullYear()
-  $('#version').text(VERSION)
-  $('#year').text(DATE)
-})
+//about.js
+
+// from 'utils.js'
+/*   global feedbackPageURL */
+const VERSION = chrome.runtime.getManifest().version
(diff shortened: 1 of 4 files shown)


scripts/about.js
```diff
@@ -1,6 +1,9 @@
...
-})
+//about.js
+
+// from 'utils.js'
+/*   global feedbackPageURL */
+const VERSION = chrome.runtime.getManifest().version
+const DATE = new Date().getFullYear()
+$('#version').text(VERSION)
+$('#year').text(DATE)
+$('#reviews-page').attr('href', feedbackPageURL)
```
anishsarangi, line 9: Add target='_blank' attribute also

tikhsuP: Thanks, was thinking of adding that in other links too.

cgorringe: I was planning to add target='_blank' to all of the links in the About page, but if you want to, go ahead.

tikhsuP: I added it in the `about.html` file. Please review : )

PR #574, opened 2020-06-22
Misc Tweaks 1

cgorringe:
Some code tweaks:
- Prevents "Outlinks" and "Screenshot" checkbox text from being highlighted.
- Removes duplicated code that initializes settings.
- Agreement page now only displays on first install and not updates.
- Updates the settings defaults to mostly True on install.
- Rearranges General Settings.

### DIFF
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -294,23 +294,25 @@ function attachTooltip (anchor, tooltip, pos = 'right', time = 200) {
     })
 }
 
-function resetExtensionStorage () {
+function initDefaultOptions () {
   chrome.storage.sync.set({
-    show_context: 'tab',
-    resource: false,
-    auto_update_context: false,
-    wm_count: false,
-    auto_archive: false,
-    email_outlinks: false,
     spn_outlinks: false,
     spn_screenshot: false,
+    /* General */
+    wm_count: true,
+    resource: true,
+    auto_archive: true,
+    email_outlinks: true,
+    not_found_popup: true,
+    auto_update_context: true,
+    show_context: 'tab',
+    /* Contexts */
+    showall: false,
     alexa: false,
     domaintools: false,
     wbmsummary: false,
     annotations: false,
-    tagcloud: false,
-    showall: false,
-    not_found_popup: true
+    tagcloud: false
   })
 }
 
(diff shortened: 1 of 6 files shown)


scripts/utils.js
```diff
@@ -294,25 +294,26 @@ function attachTooltip (anchor, tooltip, pos = 'right', time = 200) {
...
   chrome.storage.sync.set({
-    agreement: false,
-    show_context: 'tab',
-    resource: false,
-    auto_update_context: false,
-    wm_count: false,
-    auto_archive: false,
-    email_outlinks: false,
     spn_outlinks: false,
     spn_screenshot: false,
```
tikhsuP, line 300: These are 'false' but the 'Email Outlinks' option in 'General' menu is true. But to get an Email, the 'Outlinks' option should be selected, right?

scripts/utils.js
```diff
@@ -294,25 +294,26 @@ function attachTooltip (anchor, tooltip, pos = 'right', time = 200) {
...
-    showall: false,
+    selectedFeature: null,
+    /* General */
+    wm_count: true,
+    resource: true,
+    auto_archive: true,
+    email_outlinks: true,
     not_found_popup: true,
-    selectedFeature: null
+    auto_update_context: false,
```
tikhsuP, line 308: Should we make it 'true', just because all other options are selected by default?

tikhsuP: I can push these changes if these are needed : )

scripts/utils.js
cgorringe: Outlinks (if checked in Settings) will only be emailed if the Outlinks checkbox under the SPN button is also checked. We don't want those checked by default because Outlinks takes longer for the wayback machine to save, so want them only if users select them.

scripts/utils.js
cgorringe: I initially tried it as true, but since it'll update Context every time a user switches tabs or opens a new page, I felt it would be too much activity going on if it were the default. I suppose I could go either way on this.

tikhsuP: Okay, approving the changes : )

scripts/utils.js
tikhsuP: Thanks for explaining.

scripts/utils.js
```diff
@@ -294,25 +294,26 @@ function attachTooltip (anchor, tooltip, pos = 'right', time = 200) {
     })
 }
 
-function resetExtensionStorage () {
+function initDefaultOptions () {
   chrome.storage.sync.set({
```
anishsarangi, line 298: When you remove the extension from firefox and again install it, the welcome page will not be shown.

cgorringe:
Ok, fixed the issue with Firefox.
Also turned off the Settings by default first, then turns them on only after Agreement.

PR #579, opened 2020-06-24
Change storage to local and Use arrow functions

tikhsuP:
This PR changes the storage type from `chrome.storage.sync` to `chrome.storage.local` and upgrades the anonymous functions to arrow functions format.


cgorringe:
there are merge conflicts

PR #583, opened 2020-06-26
Moving files into webextension directory

cgorringe:
Mostly moving files, updating docs and config files.
- Already merged in PR #582.
- Made one minor fix to pass `npm test`.
- Haven't tested `webpack.config.js` and it's possible that I got the paths wrong, so please test by reading first, then running webpack.


tikhsuP:
Looks good!

anishsarangi:
Webpack configuration looks good after the change

PR #584, opened 2020-06-26
Improve working of Social Share section

tikhsuP:
This PR improves functioning of the social share section. Now the sharing options work when clicked on the button or brand logo (fb/twitter/linkedin).


tikhsuP:
The message 'This page was never archived' was not getting displayed for the pages that were never saved. Slightly changing the if statement is making it work.

anishsarangi:
This improvement also works fine !

cgorringe:
I'm looking deeper into that 1 line change in showWaybackCount, so will report in a bit!

cgorringe:
I added a bit of code, as I thought to include this improvement to sharing on social media. Before it used to only share a wayback machine line to the Overview page. Now what it does is include a timestamp of the current time in the URL which will automatically display the last known snapshot prior to that timestamp. This is better than using the generic "most recent" URL because that may bring up a different snapshot in the future, while this URL will always point to the same snapshot in time.

PR #585, opened 2020-06-26
Removed storing of newshosts array in storage

anishsarangi:
In this PR I have used utils.js for storing newshots array rather than storing in chrome storage


anishsarangi:
Thank you !!

PR #593, opened 2020-07-02
Re-design Context Page

tikhsuP:
This PR provides a new look to the context page.


webextension/css/context.css
```diff
@@ -1,19 +1,52 @@
...
+    color: #fff;
+}
+
+.ia-logo {
+    height: 64px;
+}
+
+.wb-logo {
+    height: 100px;
 }
```
cgorringe, line 50: I'd change this to:
`.wb-logo { width: 100%; padding-top: 1em; }`

If you want me to, I can modify the logo image to make the white area transparent so it looks good in Dark Mode. (or create a new image...)

cgorringe:
Also the <h1> title would look better if it matched the same used on the About page.

webextension/css/context.css
tikhsuP: Yes, please make the logo image transparent. That would help for the dark-mode.

webextension/css/context.css
```diff
@@ -142,14 +142,16 @@ ul {
...
 
 .pre {
-    /* font-family: Menlo,Monaco,Consolas,"Courier New",monospace; */
+    font-family: Menlo,Monaco,Consolas,"Courier New",monospace;
     display: block;
-    padding: 9.5px;
-    margin: 1em 0 5vh;
```
tikhsuP, line 102: The margin issue is because of removing it, we can retain/add it :-)

anishsarangi: If you will add here there is issue in overview context page !

anishsarangi: I have made the changes in latest changes, you can check !

tikhsuP: Yes, thank you !

cgorringe:
- The border & background colors of the frames in Alexa and Wayback should match that in Annotations. (gray border instead of white)
- Overall Menu border could have rounded corners, same as other borders in bodies.
- Bottom of Domain box in Annotations needs 1em padding or margin on bottom.
- Try Show Contexts in a Window resized to half the screen width, then reduce the font sizes overall in the body text (h2 title is fine) and possibly reduce the menu font size just slightly.

cgorringe:
It is getting better, and it's looking more responsive when resizing the window! I wish it were more consistent and matched the style of the toolbar popup. So this is what I'd like to see:

- Font size in body is still too large for Alexa, Wayback, and Annotations.
- Annotations page: Reduce space between "URL" & container, "Domain" & container.
- Keep spacing the same above & below title for the first 3, similar to the Tag Cloud page.
- Use the same overall background color in the body as the Popup in both Light and Dark modes. Darken or lighten slightly for the left column background and may need to tweak the background inside the containers.
- While the body container border color look nice, let's change it to match the border colors used in Popup Settings. (and see how that looks, or we may decide to change it back...)
- We should redo the menu to match the (General | Contexts) buttons from Settings to keep it consistent. (light and dark modes) Yeah this is a big change here.

tikhsuP:
I made the mentioned changes and I would like some suggestions on adding a border to the Menu.
The pushed code looks more similar to the General | Contexts buttons in the popup, and the one shown below is a suggestion, we can also match this Menu border color to that of body container.

Capture11

Capture21

anishsarangi:
I would prefer to go with the suggested changes ! 🙂

cgorringe:
I pushed some color tweaks and fixed the tabs border-radius. Everything looks great now as far as color and spacing! Don't need to add a border to the menu. The only thing left is if you can reduce the font size of main text in the body (title is fine). Now that I can see the wayback logo, I may tweak that a little as well, but I can do that later in a separate branch.

webextension/css/context.css
```diff
@@ -43,6 +43,14 @@ h4 {
     color: #333;
 }
 
+.sidebar-tabs div:first-child button {
+    border-radius: 5px 5px 0 0 !important;
+}
+.sidebar-tabs div:last-child button {
+    border-radius: 0 0 5px 5px !important;
+}
```
tikhsuP: Thank you for this @cgorringe 
This always adds the borders to the first and last tabs (Alexa and Tag Cloud) written in HTML markup, irrespective of the first and last tab that is displayed in the Menu. I made some JS modifications to achieve the desired behavior. Please check :-)

PR #597, opened 2020-07-06
Add private-mode feature

anishsarangi:


anishsarangi:
@tikhsuP The work is still in progress ! And I intentionally kept "404 popup" checked because this is our main feature .
Thank you

cgorringe:
In this case let's turn off the 404 feature when Private Mode is ON. :-)

anishsarangi:
Ok. I will turn it off when the private-mode is ON !

anishsarangi:
@tikhsuP 
1. Now I am working on the API call part. It might take some time !
2. Nice catch. Thank you! I will fix that too 👍

webextension/scripts/settings.js
```diff
@@ -146,6 +214,7 @@ function switchTabWindow() { $('input[type="radio"]').not(':checked').prop('chec
 
 function addDocs () {
   let docs = {
+    'private-mode': 'Extension will not support any feature!',
```
tikhsuP, line 217: It looks good!
But, when I am not logged in and I hover over the SPN button (Private Mode OFF), it should disable the SPN button and flipped button should say "Login to save", right? but it doesn't, instead it shows me the last save.

tikhsuP: And, should we change the tooltip text?  :-)

anishsarangi: Ok. Let me look into this . Thank you !

cgorringe:
Haven't studied the code yet, but it's still calling the Availability API when user taps on the toolbar icon, so that will need to be solved.

webextension/scripts/background.js
```diff
@@ -350,12 +344,13 @@ chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
...
   } else if (message.message === 'auth_check') {
-    if (!isLoggedIn) {
-      auth_check()
-      .then(resp => sendResponse(resp))
-    } else {
-      sendResponse(isLoggedIn)
-    }
+    chrome.cookies.get({ url: 'https://archive.org',name: 'logged-in-sig' }, (result) => {
+      if(result!== null){
+        sendResponse(true)
```
cgorringe, line 332: `sendResponse(result !== null)` // one liner :-)

webextension/scripts/settings.js
```diff
@@ -88,6 +128,52 @@ function selectall () {
   }
 }
 
+function validatePrivateMode (event) {
+  let checkboxes = $('[name="private-inlude"]')
```
cgorringe, line 132: spelling error: `private-inlude`

webextension/scripts/settings.js
```diff
@@ -88,6 +128,52 @@ function selectall () {
...
+        $(event.target).addClass('selected-prior')
+      }
+    }
+  }
+}
+
+function togglePrivateMode () {
+  if ($('#private-mode').is(':checked')) {
+    chrome.runtime.sendMessage({ message: 'clearCountBadge' })
+    chrome.runtime.sendMessage({ message: 'clearCountCache' })
```
tikhsuP, line 169: Do we need these? When I turn Private mode ON, the toolbar state and badge count are cleared even if we remove these lines.
Also, if we are clearing the Badge count and Toolbar state without needing to reload the popup then we need to hide blue buttons and "Saved X times" text too. It already works fine when I reload the popup, because the WB Count and Archived Resources options aren't selected.

Maybe we should do somehting like this:
`if ($('#private-mode').is(':checked')) {`
    `$('#wayback-count-label').hide()`
    `$('#borrow_books').hide()`
    `$('#news_recommend').hide()`
    `$('#wikibooks').hide()`
    `$('#doi').hide()`
  `}`

And, when I turn private mode OFF after reloading the popup, then reload the page, it doesn't shows 'R' and 'WB Count'. This is because it doesn't restores the previous selections after popup reload.

tikhsuP: Should we disable/enable the last save, when Private mode is ON/OFF, without needing to reload the popup?

anishsarangi: @tikhsuP Thanks for some good testing. Please test with the new changes I have pushed and let me know how I can improve it better. I am able to restore the previous state even if the popup is closed

cgorringe:
1. Checking Private Mode didn't properly clear the other Settings (like Wayback Count) and the sparkline API still got called when both Wayback Count and Private Mode were checked.
2. When checking Private Mode did work, it cleared all of the Settings. (should not clear Email Outlinks, or Show Resources During Save)
3. Clearing All of the Settings shouldn't automatically turn ON Private Mode. The user should be the only one to turn ON Private Mode.

webextension/scripts/settings.js
```diff
@@ -88,6 +99,68 @@ function selectall () {
...
+
+  // Checking previous state when the only one item is checked and remove 'selected-prior' class from everywhere
+  if (checkedCount === 0 && $('.selected-prior').length === 1) {
+    $('.selected-prior').removeClass('selected-prior')
+  }
+
+  // If the event.taget.checked is true,then check if it the checkbox is only item, then remove 'selected-prior' from all
+  // And add class 'selected-prior' to the event.taget, if it is NOT there
+  // Else If the event.taget.checked is false, check if the check-box is NOT the only item, then remove class 'selected-prior'
+  if (event.target.checked === true) {
```
cgorringe, line 119: The way some of this is being handled doesn't seem right to me, that is showing and hiding things like `#borrow_books` or `#wayback-count-label` should be written once in the code where settings are set or cleared, and the only code that should be in this function is to set or clear those settings. This may involve reorganizing code so that each setting as a function to clear and set each, but that organization would be preferable than to have code copied in this function.

cgorringe:
Something didn't work right when I initially checked Private Mode ON, and Relevant Archived Resources didn't clear, but it did later on after checking it.

cgorringe:
Still encountered this bug:
1. When I tested it out, it assumed I was logged in, and when I pressed "Save Page Now", it called /save/ API, but the API returned the message "You need to be logged in to use Save Page Now." so it looks like the check needs to account for that. (Private Mode was checked.) It didn't leave me with any way to log in.

cgorringe:
So the bug still persists regarding pressing "Save Page Now" while logged out, it'll call the SPN API, and not provide any means to login. You can recreate this issue by going to Chrome > Clear Browsing Data... and select Cookies to clear. This will force a log out.

We can't merge this to master until this is fixed, as it's a major bug.

webextension/scripts/background.js
```diff
@@ -90,19 +97,16 @@ function savePageNow(tabId, page_url, silent = false, options = []) {
   }
 }
 
-function auth_check() {
+function authCheckAPI() {
```
anishsarangi, line 100: I am not sure why @cgorringe re-added this. !

anishsarangi: This is not being used anywhere !

PR #603, opened 2020-07-07
Clean URL when on 'web.archive.org'

tikhsuP:
- [x] Return to treating web.archive.org URLs as a special case, and using the extracted URL to act on the following:

- Fetching the WB Count / Badge Count

- SPN (in the Popup as well as in the right click menu)

- Oldest/Overview/Newest (in the Popup as well as in the right click menu)

- Site Map

- Related Tweets

- Show Contexts

- Social Share

- Search Box

### DIFF
--- a/test/utils.spec.js
+++ b/test/utils.spec.js
@@ -3,6 +3,7 @@ const expect = require('chai').expect
 const assert = require('assert').strict
 const getUrlByParameter = require('../webextension/scripts/utils').getUrlByParameter
 const isValidUrl = require('../webextension/scripts/utils').isValidUrl
+const get_clean_url = require('../webextension/scripts/utils').get_clean_url
 const isNotExcludedUrl = require('../webextension/scripts/utils').isNotExcludedUrl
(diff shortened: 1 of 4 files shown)


tikhsuP:
Sorry, I accidentally deleted the 'package' files, working to restore these.

tikhsuP:
> Should we be calling `get_clean_url()` in the Search Box? (`search_box_activate()`) Seems pretty rare for the web.archive.org URL to occur there, then again, it's probably fine to call, so I can't really tell one way or the other.

@cgorringe  Yes, you are right, this is a rare case but if we do not call `get_clean_url()` in the Search Box (`search_box_activate()`) then the user will be taken to [https://web.archive.org/web/* /https://web.archive.org/web/ */https://www.google.com/](https://web.archive.org/web/*/https://web.archive.org/web/*/https://www.google.com/) if [https://web.archive.org/web/*/https://www.google.com/](https://web.archive.org/web/*/https://www.google.com/) is entered inside the searchbox :-)

PR #607, opened 2020-07-09
Use search box value for extension features

tikhsuP:
- [x]  When the Search Box isn't empty, use that domain as the URL to act on instead of the web browser's address bar.

- Also Hide the WB Count and Archived Resources (`'R'` icon and blue buttons), if these exist.

### DIFF
--- a/webextension/scripts/popup.js
+++ b/webextension/scripts/popup.js
@@ -1,7 +1,7 @@
 // popup.js
 
 // from 'utils.js'
-/*   global isValidUrl, isNotExcludedUrl, openByWindowSetting, hostURL, feedbackPageURL, newshosts, dateToTimestamp */
+/*   global isValidUrl, isNotExcludedUrl, openByWindowSetting, hostURL, feedbackPageURL, newshosts, dateToTimestamp, searchValue */
 
 function homepage() {
   openByWindowSetting('https://web.archive.org/')
@@ -9,7 +9,7 @@ function homepage() {
 
 function save_now() {
   chrome.tabs.query({ active: true, currentWindow: true }, (tabs) => {
-    let url = tabs[0].url
+    let url = searchValue.length >= 3 ? searchValue : tabs[0].url
     let options = ['capture_all']
     if ($('#chk-outlinks').prop('checked') === true) {
       options.push('capture_outlinks')
@@ -43,7 +43,7 @@ function last_save() {
     } else {
       $('#save_now').removeAttr('disabled')
       chrome.tabs.query({ active: true, currentWindow: true }, (tabs) => {
-        let url = tabs[0].url
+        let url = searchValue.length >= 3 ? searchValue : tabs[0].url
         chrome.runtime.sendMessage({
           message: 'getLastSaveTime',
           page_url: url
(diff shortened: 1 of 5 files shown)


webextension/scripts/popup.js
```diff
@@ -1,15 +1,15 @@
...
+/*   global isValidUrl, isNotExcludedUrl, openByWindowSetting, hostURL, feedbackPageURL, newshosts, dateToTimestamp, searchValue */
 
 function homepage() {
   openByWindowSetting('https://web.archive.org/')
 }
 
 function save_now() {
   chrome.tabs.query({ active: true, currentWindow: true }, (tabs) => {
-    let url = tabs[0].url
+    let url = searchValue.length >= 3 ? searchValue : tabs[0].url
```
cgorringe, line 12: '3' could be a const assigned at the start of the file.

tikhsuP: I found a way to make it work without using it.
Thank you!

webextension/scripts/popup.js
```diff
@@ -139,15 +139,28 @@ function search_tweet() {
   })
 }
 
+function makeValidURL() {
+  if (!isValidUrl(searchValue)) {
+    searchValue = 'https://' + searchValue
+  }
+}
+
```
cgorringe, line 147: Instead of using a global, better to pass in the variable to this function?

webextension/scripts/popup.js
```diff
@@ -147,16 +147,27 @@ function makeValidURL() {
 
 function useSearchBoxValue() {
```
cgorringe, line 148: Would be better if you can make this work by passing in seachValue into the function, rather than relying on a global seachValue.

webextension/scripts/popup.js
```diff
@@ -139,11 +139,47 @@ function search_tweet() {
   })
 }
 
+function makeValidURL() {
+  if (!isValidUrl(searchValue)) {
+    searchValue = 'https://' + searchValue
+  }
+}
+
+function useSearchBoxValue() {
```
cgorringe, line 148: Would be better if you can make this work by passing `searchValue` in to the function, rather than referencing it as a global. I think you can also change the code to avoid the recursive call to `useSearchBoxValue()` by moving the call to `makeValidURL()` to the front of the function (which should also return the valid URL instead of modifying a global.)

webextension/scripts/popup.js
cgorringe: Nope. That's not the way to do it. Here:

```suggestion
function makeValidURL(url) {
  return isValidUrl(url) ? url : 'https://' + url
}

```

webextension/scripts/popup.js
tikhsuP: Sorry, I misinterpreted the line above and moved the call to `useSearchBoxValue()` to `makeValidURL()` and couldn't find a way to return the value. Sorry.

Now I have moved the call to `makeValidURL()` to the desired place.

webextension/scripts/popup.js
tikhsuP: Thank you!

tikhsuP:
I resolved the conflicts and added `get_clean_url()` for search box value (searchValue). Thank you!

webextension/scripts/utils.js
```diff
@@ -176,7 +176,7 @@ function remove_wbm(url) {
 
 // Function to clean the URL if the user is on 'web.archive.org'
 function get_clean_url(url) {
-  if (url.includes('web.archive.org')) {
+  if (url && url.includes('web.archive.org')) {
```
tikhsuP, line 179: Thank you for adding `url &&` to the if-condition statement. I tried `get_clean_url(makeValidURL(event.target.innerHTML))` earlier but was getting `Cannot read property 'includes' of null` earlier, in case when `null` was returned :-)

PR #608, opened 2020-07-12
Updates to support in Microsoft Edge

tikhsuP:
- Added excluded URLs for edge.
- Updated README.md.

### DIFF
--- a/webextension/scripts/utils.js
+++ b/webextension/scripts/utils.js
@@ -3,7 +3,7 @@
 let isArray = (a) => (!!a) && (a.constructor === Array)
 let isObject = (a) => (!!a) && (a.constructor === Object)
 
-let isFirefox = (navigator.userAgent.indexOf('Firefox') > -1)
+let isFirefox = (navigator.userAgent.indexOf('Firefox') !== -1)
 const hostURL = isFirefox ? 'https://firefox-api.archive.org/' : 'https://chrome-api.archive.org/'
 const feedbackPageURL = isFirefox ? 'https://addons.mozilla.org/en-US/firefox/addon/wayback-machine_new/' : 'https://chrome.google.com/webstore/detail/wayback-machine/fpnmgdkabkmnadcjpehmlllkndpkmiak/reviews?hl=en'
 
@@ -134,7 +134,8 @@ const excluded_urls = [
   '192.168.',
   '10.',
   'file:',
-  'edge:'
+  'edge:',
+  'extension:'
 ]
 
 // Function to check whether it is a valid URL or not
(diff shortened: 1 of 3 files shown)


webextension/scripts/utils.js
```diff
@@ -3,7 +3,7 @@
 let isArray = (a) => (!!a) && (a.constructor === Array)
 let isObject = (a) => (!!a) && (a.constructor === Object)
 
-let isFirefox = (navigator.userAgent.indexOf('Firefox') !== -1)
+let isFirefox = (navigator.userAgent.indexOf('Firefox') > -1)
```
anishsarangi, line 6: Why did you change this ?

webextension/scripts/utils.js
```diff
@@ -133,7 +133,8 @@ const excluded_urls = [
   'moz-extension:',
   '192.168.',
   '10.',
-  'file:'
+  'file:',
+  'edge:'
```
anishsarangi, line 137: You also can add a new case in `isValidUrl` and `isNotExcludedUrl` in utils.spec.js

webextension/scripts/utils.js
tikhsuP: I was checking out MDN web docs documents and I found this: https://developer.mozilla.org/en-US/docs/Web/API/Window/navigator
They used ` > -1` so, I made the change.

webextension/scripts/utils.js
tikhsuP: Okay, will do it. Thanks :-)

webextension/scripts/utils.js
tikhsuP: Would you like me to revert it? :-)

anishsarangi: No. You may create another variable and check for edge browser. Please ask Owenl to create new host name for edge browser. !
@cgorringe may need to check this once ! Thank you !

tikhsuP: Yes, we need `edge-api`, and I will add the check for that. Thank you!

cgorringe: We may want to rethink how we use isFirefox etc... as we'll have multiple browsers to support, including Edge and Safari.

Using `!== -1` vs `> -1` doesn't really make any difference. I'd leave it at `!== -1`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf

webextension/scripts/utils.js
tikhsuP: I added the tests. Thank you!

webextension/scripts/utils.js
tikhsuP: I reverted this change. Thank you!

PR #614, opened 2020-07-16
Use utils function to get the overview count

anishsarangi:
In this PR I have improved the overview context

### DIFF
--- a/webextension/scripts/overview.js
+++ b/webextension/scripts/overview.js
@@ -28,7 +28,7 @@ function get_details () {
     if (captures > 0) {
       $('#total_captures').show()
       get_thumbnail()
-    }else{
+    } else {
       $('#loader_thumbnail').hide()
       $('#show_thumbnail').text('Thumbnail not found.')
     }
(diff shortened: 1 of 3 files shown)


cgorringe:
Looks like `timestamp2datetime()` and `_splitTimestamp()` in `overview.js` can be replaced by calling `timestampToDate()` in `utils.js` (which I wrote later, having not noticed these were already written...)

webextension/scripts/overview.js
```diff
@@ -1,40 +1,38 @@
...
-        .text(getTotal(captures).toLocaleString())
-    } else {
-      $('#total_archives_number').text(0)
     }
   }).fail(() => {})
+  var captures
+  getWaybackCount(url, (total) => {
+    captures = total
+    $('#total_archives_number').attr('href', 'https://web.archive.org/web/*/' + url)
+    .text(captures)
```
cgorringe, line 27: if you add `.toLocaleString()` to `captures` it'll add commas to the number!

PR #618, opened 2020-07-18
Update README.md

JHBalaji:
fixed chrome store url.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Follow the steps below for the currently deployed version.
 
 ### Chrome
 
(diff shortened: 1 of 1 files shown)


cgorringe:
I just closed the same PR less than 24 hours ago. We're going back to the prior ID and URL, it's just taking a while to get it back up!

We'll keep this PR open so that others can see this until we get the old URL restored.

JHBalaji:
No problem. Thanks!

cgorringe:
25 days later. Restored. Finally!

PR #623, opened 2020-07-19
Fix Context screen bugs

anishsarangi:
This PR fixes the bugs on Context pages:

- [x]  Alexa Overview should only work on the domain name. Anything after should be stripped.
- [x] When passing the url value on the internal context.html URL, the url needs to be encoded to support ? characters.
- [x] Check to make sure Wayback Captures Overview works correctly on URLs that are not in the archive, and display an appropriate message. Could hide the usual message and display one that says that the URL has not been archived.


cgorringe:
First 2 looks good!

PR #624, opened 2020-07-19
Add Bulk Save Option

tikhsuP:
This PR adds an option to Bulk Save, which:
- [x] 1. Imports bookmarks
- [x] 2. Adds and save multiple URLs at once
- [x] 3. Show the "Bulk Save" button only if the user is logged in.
- [x] 4. Cancel the saves.
- [x] 5. Saves 5 URLs at once, then add more as save completes.
- [x]  Use textarea instead of input box.


webextension/scripts/bulk-save.js
```diff
@@ -0,0 +1,170 @@
...
+      urlList.add(node.url)
+      displayList(urlList)
+    }
+  }
+}
+
+// import all bookmarked URLs
+function importBookmarks() {
+  $('#empty-list-err').hide()
+  chrome.bookmarks.getTree((itemTree) => {
```
cgorringe, line 56: I got this error after tapping "Import All Bookmarks":
`Uncaught TypeError: Cannot read property 'getTree' of undefined`

tikhsuP: Ok, thank you. I will remove this error.

cgorringe:
A lot of this works well, Good Job!

I was thinking when the Auth check to see if user is logged-in is done in the main popup (for the "Save Page Now" button) you can show or hide the "Bulk Save" button, then you wouldn't need to do it again for the "Save All" button in the Bulk Save window. 

If, while saving each URL in the list, it turns out the user wasn't actually logged in (or they're offline) then you'll want to cancel the rest of the saves from occurring.

Right now since you call initiateBulkSave() for every item in the list all at once without waiting for any priors to complete, the API will likely start returning errors about too many concurrent saves happening at once. But this is a good start. Now we just need to figure out the next step of limiting to 5 saves at a time, then adding more as those complete.

webextension/scripts/bulk-save.js
cgorringe: I'm not sure why, as it later worked correctly. Although I was logged out when this occurred, then was logged in when it worked correctly, but I wouldn't think that would matter when trying to read the bookmarks list. Or maybe I just forgot to reload the extension initially and it needed to refresh the manifest file to gain access to bookmarks. (since new permission was added.)

tikhsuP:
Saves 5 URLs when all URLs are saved, but some bugs when `wmAvailabilityCheck()` is called. So, I still need to figure out how to handle that.

cgorringe:
"Bulk Save" button shows on blank pages (URL not supported). Could do the reverse, hide by default in CSS, then call .show() if logged in?

webextension/scripts/bulk-save.js
cgorringe: I got the error again, but then refreshed the extension and it worked. Good idea to at least check that `chrome.bookmarks` isn't null.

tikhsuP: Thank you. Fixed.

cgorringe:
I like how I can copy & paste a list of URLs now to add to the list, then remove individual URLs. This is coming along nicely!

Some tweaks to make:

1. Text Area in Dark Mode

2. Avoid re-adding URLs that are already in the list.

3. One harder situation is where I added `google.com` (which added `https://google.com`) then Imported my bookmarks which added `https://www.google.com` which is technically not a duplicate, but realistically it is. So, in the duplicate-checking function, something to make it a little smarter could be to:
```
isDuplicateURL(url1, url2) {
  1. Take part after http(s) from url1 and url2.
  2. Remove the www. prefix from both.
  3. Compare for equality.
  return true or false
}

```

cgorringe:
Also noticed the the "Bulk Save" button doesn't show if you're on a "URL not supported" page such as an empty tab. Which makes sense that the Save Page Now button wouldn't do anything on that page, but Bulk Save should still work.

tikhsuP:
> Also noticed the the "Bulk Save" button doesn't show if you're on a "URL not supported" page such as an empty tab. Which makes sense that the Save Page Now button wouldn't do anything on that page, but Bulk Save should still work.

Working on this

cgorringe:
- I found one issue where if "Show Resources During Save" is checked in Settings, then a bunch of those windows will popup, one for each URL during bulk save. Since this is really annoying, let's prevent these windows from popping up during Bulk Save.
- Otherwise, since this is mostly working now, I approve and let's get this merged to master! Any remaining tweaks (such as moving buttons) we can do later.

webextension/scripts/background.js
```diff
@@ -391,12 +391,13 @@ chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
     var wayback_url = message.wayback_url
     var url = page_url.replace(/https:\/\/web\.archive\.org\/web\/(.+?)\//g, '')
     var open_url = wayback_url + encodeURI(url)
+    let isBulkSave = message.isBulkSave
```
anishsarangi, line 394: Please make it like `message.isBulkSave || false`

cgorringe:
After making Anish's suggested change and fixing the merge conflict in popup.js, I think it's good to go.

webextension/scripts/background.js
tikhsuP: I made the changes. Thank you!

PR #625, opened 2020-07-20
Exclude UP & DOWN keys from event

tikhsuP:
1. Fixes [this](https://github.com/internetarchive/wayback-machine-chrome/pull/619#pullrequestreview-451174366), enabling the user to go further down the search-box list.
2. Hides "Using Search URL" text when not using search box url.

### DIFF
--- a/webextension/css/doi.css
+++ b/webextension/css/doi.css
@@ -9,7 +9,7 @@
 #container-whole-doi > div {
     background-color: #eee;
     border-radius: 5px;
-    border: 1px solid black;
+    border: 1px solid #888;
     margin: 10px auto;
     padding: 5px;
     justify-content: space-between;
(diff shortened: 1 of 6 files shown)


cgorringe:
The up & down keys work now, but I'm also seeing that the list box oftentimes disappears entirely. For example it'll show as I type `google` but then disappears after including the period on `google.` or `google.com` (adding the period also displays "Using Search URL" which is fine.)

anishsarangi:
@cgorringe I think that is fixed in #619.

tikhsuP:
The long link text used to go beyond the desired DOM, like this:

Screenshot (137)

I have fixed this by pushing the text to another line.

tikhsuP:
I have merged 'master' and resolved conflicts.

PR #626, opened 2020-07-20
Add API for Edge

tikhsuP:
This PR adds API for Microsoft Edge Browser.

### DIFF
--- a/webextension/scripts/utils.js
+++ b/webextension/scripts/utils.js
@@ -5,8 +5,23 @@ let isObject = (a) => (!!a) && (a.constructor === Object)
 
 let searchValue
 let isFirefox = (navigator.userAgent.indexOf('Firefox') !== -1)
-const hostURL = isFirefox ? 'https://firefox-api.archive.org/' : (navigator.userAgent.indexOf('Edg') !== -1 ? 'https://edge-api.archive.org/' : 'https://chrome-api.archive.org/')
-const feedbackPageURL = isFirefox ? 'https://addons.mozilla.org/en-US/firefox/addon/wayback-machine_new/' : 'https://chrome.google.com/webstore/detail/wayback-machine/fpnmgdkabkmnadcjpehmlllkndpkmiak/reviews?hl=en'
+let isEdge = (navigator.userAgent.indexOf('Edg') !== -1)
+
+const hosts = {
+  firefox: 'https://firefox-api.archive.org/',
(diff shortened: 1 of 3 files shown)


cgorringe:
It looks fine. I think I'll also push a test for Safari to this PR in a bit.

cgorringe:
Updated to detect most major browsers.

webextension/scripts/utils.js
```diff
@@ -23,6 +19,50 @@ const newshosts = new Set([
   'www.washingtonpost.com'
 ])
 
+/* * * Browser Detection * * */
+
+function getBrowser() {
+  // the order of these is important!
+  if (navigator.brave) { return 'brave' }
+  else if (navigator.userAgent.indexOf('Edg') !== -1) { return 'edge' }
```
tikhsuP, line 27: Should we detect `indexOf( )` for both 'Edge' (Legacy Version) & 'Edg' (Chromium verison)?

cgorringe: Yes, you're correct.

webextension/scripts/utils.js
```diff
@@ -23,6 +19,50 @@ const newshosts = new Set([
...
+
+const feedbackURLs = {
+  chrome: 'https://chrome.google.com/webstore/detail/wayback-machine/fpnmgdkabkmnadcjpehmlllkndpkmiak/reviews?hl=en',
+  chromium: 'https://chrome.google.com/webstore/detail/wayback-machine/fpnmgdkabkmnadcjpehmlllkndpkmiak/reviews?hl=en',
+  firefox: 'https://addons.mozilla.org/en-US/firefox/addon/wayback-machine_new/',
+  safari: 'https://apps.apple.com/us/app/wayback-machine/id1201888313'
+}
+
+const gBrowser = getBrowser()
+const isChrome = (gBrowser === 'chrome') || (gBrowser === 'chromium')
```
anishsarangi, line 56: Do we really need these variables (except isFirefox) at this point of time ?

cgorringe: I'm going to need them in the future, possibly. I'll have to make some changes like converting all `chrome` to `browser` and there are some browser-specific issues, so to get it all working with Safari as well. So let's keep it for now.

anishsarangi: Ok. Thank you. !

PR #629, opened 2020-07-23
Show 'R' only when resources are available

### LINKED ISSUE #602: Show "R" icon only if relevant resources are actually available for current URL
The add-on icon is currently displaying an "R" (promising "Relevant Archived Resources") for all URLs on the domains where such resources _might_ be available. The extension should instead check if there is anything available for the given URL, and otherwise not display the "R".
Wayback add-on  R  icon Screenshot from 2020-07-07

(Example: I tried several pages on vox.com, such as [this URL](https://www.vox.com/2020/7/6/21314472/covid-19-coronavirus-us-cases-deaths-trends-wtf), which were all showing a blue button promising "Related TV Clips" that however resulted in "Error: Clips not found" every time.)

_(as discussed on Slack)_

tikhsuP:
This PR Fixes #602 and shows 'R' only when resources are available.


cgorringe:
It "works" but it also makes a repeated API calls. For every webpage visited, calls the Papers API once and the Books API 2 times (the second call has more of the URL encoded.) It also retrieved the book cover images for Cited Books (this was before ever tapping the Cited Books button) So it looks like we need to figure out if we can reduce the API calls.

tikhsuP:
I reduced the API calls for wikipediaBooks and tvClips by storing the data in cache. This has also reduced the loading time of the related html pages (like booklist.html) in our extension.

cgorringe:
Just testing it out on Wikipedia, let's say I go to this link:
https://en.wikipedia.org/wiki/World_War_II
This will GET this API 2 times:
https://chrome-api.archive.org/services/context/books?url=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FWorld_War_II
The toolbar icon doesn't turn 'R' yet, then when I hit the reload button, it'll GET the above API 1 more time, and finally display the 'R' icon.

webextension/scripts/background.js
```diff
@@ -256,24 +256,32 @@ function getTvNews(Url, onSuccess, onFail) {
...
 function getCachedWikipediaBooks(url, onSuccess, onFail) {
-  let cacheWikipediaBooks = wikipediaBooksCache[url]
+  let cacheWikipediaBooks = wikipediaBooksCache.get(url)
   if (cacheWikipediaBooks) {
     onSuccess(cacheWikipediaBooks)
   } else {
     getWikipediaBooks(url, (json) => {
-      wikipediaBooksCache[url] = json
+      if (wikipediaBooksCache.size > 2) {
+        let first_key = wikipediaBooksCache.entries().next().value[0]
```
cgorringe, line 265: This may also work: `let first_key = wikipediaBooksCache.keys().next().value`

anishsarangi: Thank you. I will make the change.

cgorringe:
This time I was on Wikipedia, and the Books API got called 2 times on a page, then again 1 time when I clicked "Find Sited Books".

cgorringe:
I thought the code looked alright, but it's not showing the 'R' icon when it's supposed to when I tested, and stopped making API calls when it should.

cgorringe:
I made some substantial changes, and it tested fine for me. But if anyone wants to test this some more, it could use some QA. We're testing Resources so anything with Wikipedia, TV News, Amazon Books.

anishsarangi:
Thank you @cgorringe . I will have a look at it.

cgorringe:
I did once encounter the Books API being called twice, but wasn't able to duplicate the issue. Later tests only called it once and the cache working correctly.

One BUG that I did find was that if there are No Books (e.g. ISBN IDs) on the page, but there are Papers (e.g. "doi:" IDs) on a Wikipedia page, the 'R' icon will not show, and as a result the 'Cited Papers' button will not display.

I think we may want to rework this in the future. We can scan for ISBN IDs for Books, and DOI (and others) for Papers, without calling an API, as an initial step perhaps before calling the API. The 'Cited Books' and 'Cited Papers' buttons currently either both display or not at the same time, but it's not always the case that both would be present, so we may want to only display one or the other.

Going to merge this to get it in the main branch and create another issue for the above.

cgorringe (approved):

PR #630, opened 2020-07-23
Integrate Login functionalities inside extension.

anishsarangi:


cgorringe:
Looks sweet!
Mostly works. Just need it to show when login is a success!

The `/account/login` API should return the following JSON on success:
`{
	"status": "ok",
	"message": "Successful login.",
	"referer": "https:\/\/archive.org\/"
}`

webextension/scripts/login.js
```diff
@@ -0,0 +1,64 @@
...
+            $('#log_in').val('Login')
+          }else{
+            $('#logout').show()
+          }
+      })
+      .catch(e=> console.log(e))
+}
+
+function doLogout(){
+    chrome.cookies.remove({url:'https://archive.org',name:'logged-in-sig'},()=>{
```
cgorringe, line 61: Got the error `Uncaught TypeError: Cannot read property 'remove' of undefined` in the popup's console. It's probably due to the cookie being stored in a different process than the popup? I'm not sure, but maybe try moving this code to `background.js` then sending a message to call it (since can't call it directly in background process.) Or maybe it can be solved similar to `archive.js` which you can see it under "web_accessible_resources" in `manifest.json` - since that script has access to the website the user is viewing, you'd also need access in order to clear the cookie. I hope this makes sense!

webextension/scripts/login.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+    $('#log_in').val('Please Wait...')
+    const data = new URLSearchParams()
+    data.append('username', email)
+    data.append('password',password)
+    data.append('action',"login")
+    const loginPromise = new Promise((resolve, reject) => {
+        setTimeout(() => {
+          reject(new Error('timeout'))
+        }, 30000)
```
cgorringe, line 35: May reduce the timeout from 30 sec to 5 sec (5000)?

PR #631, opened 2020-07-24
Add Fact Check feature

tikhsuP:
This PR adds 'Fact Check' feature to the extension.


webextension/scripts/popup.js
```diff
@@ -387,6 +390,46 @@ function show_wikibooks() {
   })
 }
 
+function showFactCheck() {
+  chrome.tabs.query({ active: true, currentWindow: true }, (tabs) => {
+    const url = tabs[0].url
+    const tabId = tabs[0].id
+    chrome.storage.local.get(['fact_check'], (event) => {
```
tikhsuP, line 397: We are not using this (`fact_check`) in this function.

cgorringe: true - fixed.

webextension/scripts/background.js
```diff
@@ -209,6 +210,61 @@ async function validate_spn(tabId, job_id, silent = false, page_url) {
...
+
+function getCachedFactCheck(url, onSuccess, onFail) {
+  let cacheData = fact_checked_data[url]
+  if (cacheData) {
+    onSuccess(cacheData)
+  } else {
+    getFactCheck(url, (json) => {
+      // remove older cached data
+      // TODO: should use a queue (array), not an object here
+      if (Object.keys(fact_checked_data).length > 2) {
```
cgorringe, line 259: Or Map() and map.size for length

PR #632, opened 2020-07-28
Detect both Edge Legacy and Chromium version

tikhsuP:
Detect `indexOf( )` for both 'Edge' (Legacy Version) & 'Edg' (Chromium verison) as mentioned [here](https://github.com/internetarchive/wayback-machine-chrome/pull/626#discussion_r461192902)


cgorringe:
This isn't needed. The check for 'Edg' already returns true for 'Edge'.

PR #636, opened 2020-08-01
Don't show count on web.archive.org pages

tikhsuP:


cgorringe:
We once used to have `web.archive.org` in the `excluded_urls[]` list, but now we don't, which would also be a solution, but now I don't remember why we removed it from the list. Does anyone remember why we took it out?

PR #638, opened 2020-08-06
Fix README broken chrome web store link

cdosborn:


tikhsuP:
Thank you @cdosborn, for this PR.
Although, we are about to get back the old page / ID on the Chrome store, after which we'll take down the new page / ID.

PR #648, opened 2020-08-13
Use browser names dynamically, instead of using API URLs

tikhsuP:
We are currently detecting the browser and choosing API URLs dynamically, which looks like:

``` 
const hostURLs = {
   chrome: 'https://chrome-api.archive.org/',
   chromium: 'https://chrome-api.archive.org/',
   firefox: 'https://firefox-api.archive.org/',
   safari: 'https://safari-api.archive.org/',
   brave: 'https://brave-api.archive.org/',
   edge: 'https://edge-api.archive.org/',
   ie: 'https://edge-api.archive.org/',
   opera: 'https://opera-api.archive.org/'
 }
```

The API URL structure is the same for all browser, only the browser name is different, so, in this PR I have made an attempt to reduce the above to:

 ```
const hostBrowsers = {
   chrome: 'chrome',
   chromium: 'chrome',
   firefox: 'firefox',
   safari: 'safari',
   brave: 'brave',
   edge: 'edge',
   ie: 'edge',
   opera: 'opera'
 }
```

### DIFF
--- a/webextension/scripts/background.js
+++ b/webextension/scripts/background.js
@@ -4,7 +4,7 @@
 // Copyright 2016-2020, Internet Archive
 
 // from 'utils.js'
-/*   global isNotExcludedUrl, get_clean_url, isValidUrl, notify, openByWindowSetting, sleep, wmAvailabilityCheck, hostURL, isFirefox */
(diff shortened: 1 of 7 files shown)


anishsarangi:
The new code looks like there are a lot of repetition and it looks somehow complicated to understand. 
The previous code was more organised.

Issue #474, opened 2020-04-28
Slow flipping of Save Page Now button: Performance Improvement


tikhsuP:
**Describe the bug**
I am not sure if this is a valid bug or not.
When I hover over the Save Page Now button for the first time, after restarting the extension, it takes longer to flip than the next time. This may be because it takes some time to fetch the data.

**To Reproduce**
Open a page: 
1. Open the extension by clicking on the icon in the Toolbar
2. Hover over 'Save Page Now' for the first time, notice the time it takes to flip
3. Hover over 'Save Page Now' again without restarting it, notice the time it takes to flip
4. The first time takes much longer

**Expected behavior**
It should flip quickly.

**Desktop:**
 - OS: Windows 10
 - Browser: Chrome Version 81.0.4044.122

Can we do anything about it?

mitchell-frost:
Can I work on this?

tikhsuP:
Thank you @mitchell-frost for your interest. Yes, you may work on this issue.
We plan to cache the data to reduce the flipping delay. Please feel free to ask if you need any help.

mitchell-frost:
@tikhsuP I can reproduce the bug but I don't understand the data caching. How am I supposed to achieve that?

tikhsuP:
You can understand the implementation of data caching by looking into the declaration and usage of the variable [`fact_checked_data`](https://github.com/internetarchive/wayback-machine-chrome/blob/master/webextension/scripts/background.js#L19).

When the **Fact Checks** option in **General** settings is checked, the **Fact Check** feature uses the same concept to fetch and store data (if available) automatically when a page is loaded. This happens because the [`chrome.tabs.onUpdated.addListener( )`](https://github.com/internetarchive/wayback-machine-chrome/blob/master/webextension/scripts/background.js#L523
)  is fired when a page is loaded, and it calls the [`factCheckPage(tab.id, tab.url)`](https://github.com/internetarchive/wayback-machine-chrome/blob/master/webextension/scripts/background.js#L533).

More on [chrome.tabs.onUpdated.addListener( )](https://developer.chrome.com/extensions/tabs#event-onUpdated)

tikhsuP:
@mitchell-frost, the idea is to fetch and store the data when a page loads. This will make the data available to the extension before the user hovers over the **Save Page Now** button. And there won't be any need to fetch the data on hover, which causes the delay.

mitchell-frost:
@tikhsuP thank you..I am on it !

[...]

PR #650, opened 2020-08-13
save now slow flip fix

mitchell-frost:
Makes changes to fix the slow flipping of Save Page Now button. This solves the issue #474

### DIFF
--- a/webextension/scripts/background.js
+++ b/webextension/scripts/background.js
@@ -536,16 +536,27 @@ chrome.tabs.onUpdated.addListener((tabId, info, tab) => {
   if (info.status === 'complete') {
     updateWaybackCountBadge(tab.id, tab.url)
     chrome.storage.local.get(['auto_archive', 'fact_check'], (event) => {
-      // auto save page
-      if (event.auto_archive === true) {
-        auto_save(tab.id, tab.url)
-      }
       // fact check
       if (event.fact_check === true) {
         factCheckPage(tab.id, tab.url)
       }
+      // auto save page
+      if (event.auto_archive === true) {
+        auto_save(tab.id, tab.url)
+      }
     })
   } else if (info.status === 'loading') {
+    updateWaybackCountBadge(tab.id, tab.url)
+    chrome.storage.local.get(['auto_archive', 'fact_check'], (event) => {
+      // fact check
+      if (event.fact_check === true) {
+        factCheckPage(tab.id, tab.url)
+      }
+      // auto save page
+      if (event.auto_archive === true) {
+        auto_save(tab.id, tab.url)
+      }
+    })
     var received_url = tab.url
     clearToolbarState(tab.id)
     if (isNotExcludedUrl(received_url) && !received_url.includes('web.archive.org') && !(received_url.includes('alexa.com') || received_url.includes('whois.com') || received_url.includes('twitter.com') || received_url.includes('oauth'))) {
--- a/webextension/scripts/background.js
+++ b/webextension/scripts/background.js
@@ -30,7 +30,7 @@ var private_before_default = new Set([
   'not-found-popup'
 ])
 
-function loadingCacheData(tabId, page_url) {
+function loadingCacheData() {
   chrome.tabs.onUpdated.addListener((tabId, info, tab) => {
     if(info.status == 'loading') {
       chrome.tabs.query({ active: true, currentWindow: true }, (tabs) => {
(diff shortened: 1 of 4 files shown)


webextension/scripts/background.js
```diff
@@ -514,16 +514,27 @@ chrome.tabs.onUpdated.addListener((tabId, info, tab) => {
...
       // auto save page
       if (event.auto_archive === true) {
         auto_save(tab.id, tab.url)
       }
+    })
+  } else if (info.status === 'loading') {
+    updateWaybackCountBadge(tab.id, tab.url)
+    chrome.storage.local.get(['auto_archive', 'fact_check'], (event) => {
       // fact check
       if (event.fact_check === true) {
```
anishsarangi, line 530: The same code is repeated again in the "complete" state above ^! Please remove that.

webextension/scripts/background.js
```diff
@@ -514,16 +514,27 @@ chrome.tabs.onUpdated.addListener((tabId, info, tab) => {
...
         factCheckPage(tab.id, tab.url)
       }
+      // auto save page
```
anishsarangi, line 533: Same as above

webextension/scripts/background.js
mitchell-frost: Okay

webextension/scripts/background.js
mitchell-frost: Okay

anishsarangi:
@mitchell-frost This fix doesn't help in reducing the time of flipping of Save Page Now button. 
What you can do is, you can use a dict(e.g size=10) to store urls as key and their availability api response as value.
So what will happen is, it will not call the availability api again if you click on the icon. Rather it will search for the result in the dict itself.

mitchell-frost:
Thank you. I will implement this but I don't understand which URLs I am supposed to have in the dictionary. Could you elaborate a little?

anishsarangi:
Whenever you click on the icon, the extension makes a background api call to fetch its availability. You can store that in a dict/Map in background like `{ url_name : availability_data }` . Then if you will click on the icon again after some time, it will not call the api. Rather it can get the value from Map which in fact will reduce the time. You can limit the Map size upto 10. 
I saw @tikhsuP has already mentioned about this in the comments https://github.com/internetarchive/wayback-machine-chrome/issues/474#issuecomment-672559902. 
Please check that. Thank you !

mitchell-frost:
I came across some problems while implementing the above. I first created a map and wrote a function that stores the values from URL's last saved date as the key. You can review it [https://github.com/mitchell-frost/wayback-machine-chrome/blob/save-now-slow-flip-fix/webextension/scripts/background.js#L60](here). 

In the loginSuccess() function, it iterates over the map cached_url_data instead of making the API call, chrome.storage.local.get:
[https://github.com/mitchell-frost/wayback-machine-chrome/blob/save-now-slow-flip-fix/webextension/scripts/popup.js#L75-L87](here)

The Save Now button won't flip for any URL except "URL not supported." It also doesn't show the "Archiving URL.." progress bar while saving. I don't understand where I am going wrong. @anishsarangi I shall be grateful if you could have a look and tell me where the extension is facing thee problem.

webextension/scripts/popup.js
```diff
@@ -74,7 +74,18 @@ function loginSuccess() {
       if (isNotExcludedUrl(url)) {
         $('#savebtn').click(save_now)
         $('#contextTip').click(openContextMenu)
-        chrome.storage.local.get(['private_mode'], (event) => {
+        cached_url_data.forEach((value, key) => {
+          if(value === url) {
+            if(key === 'last_save') {
+              if($('#last_save').text !== 'URL not supported') {
+                $('#last_save').text(key);
+                break;
```
tikhsuP, line 82: While executing your code, I got this error in the browser console: `Uncaught SyntaxError: Illegal break statement`.  [More on break statement.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/break)
This error is disorienting the DOM of the extension. You can see the following screenshot:

mitchell-frost

**How to check for such errors?** 
For Chrome: You can right-click the extension icon in the browser toolbar and click on the "Inspect popup" option. This will open a developer tools window.

tikhsuP: Also, the variable `cached_url_data` is defined in `background.js` but is also being used in `popup.js`, where it can't be accessed.

mitchell-frost: Thank You. I'll change the code accordingly, @tikhsuP

anishsarangi:
The issue has been solved in #692 . Hence closing this PR. Thank you !

PR #655, opened 2020-08-20
HTML/CSS Refactoring

cgorringe:
Some improvements include:
- rearranging buttons
- widening popup
- cleaning up & refactoring CSS & HTML
- renaming CSS IDs and updating JS
- moving inline styles to css files


tikhsuP (approved):
Looks good!

PR #656, opened 2020-08-20
Handle error for TV News Clips

tikhsuP:
This PR handles error for TV News Clips when JSON format is invalid.


anishsarangi (approved):
Looks good !

PR #661, opened 2020-08-24
Open window without removing toolbar icon

tikhsuP:
This PR removes focus from the newly opened window(s) and prevents the removal of the toolbar icon.


cgorringe:
While it does prevent the 'R' or 'F' icon from disappearing, it also causes the window to popup behind the main browser window, which is undesirable. So I think we need to search further for a solution.

tikhsuP (commented):
Good fix. Looks great! Thank you.
I approve.

anishsarangi (approved):

anishsarangi (approved):

PR #662, opened 2020-08-25
Fix popup height in Edge

tikhsuP:
This PR fixes the popup height in Edge.


cgorringe:
Thanks for finding this bug. Fixed in #663

Issue #475, opened 2020-04-29
Mentioning the Date/Time for the Oldest and Newest version of saved webpage.


sanyam99ag:
**Is your feature request related to a problem/ issue? Please describe.**
According to the updated repository the page save status is being displayed on the flipped "Save Page Now" Button.
The feature that can be added is we can mention the respective Date/Time on the flipped "Oldest" and "Newest" Button. So the user will get to know about the oldest and the newest version of the webpage without checking that explicitly.

**Describe the solution you'd like**
It would be helpful to the user to show the respective Date/Time on the flipped "Oldest" and "Newest" Buttons

tikhsuP:
That would be great.

cgorringe:
Not sure if it'll fit well on the smaller buttons. I guess someone can try it and see.

sanyam99ag:
@cgorringe Surely the date will fit in the button if we use the "YY/MM/DD" format.
&nbsp;  &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;        ![Oldest](https://user-images.githubusercontent.com/39384430/80859191-5221c680-8c7c-11ea-8fc8-f142ee824332.png) &nbsp;  &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;    ![Newest](https://user-images.githubusercontent.com/39384430/80859189-50580300-8c7c-11ea-8f21-f8d9ec858ec2.png)

tikhsuP:
I think we should follow the same pattern as in the Save Page Now button, as using a different pattern will confuse the user.
Also, the YYYY/MM/DD is the easiest to understand. The patterns like YY/MM/DD, DD/MM/YY, MM/DD/YY are more probable to be misunderstood.

Can we try a popup message with YYYY-MM-DD that shows up on hover?

sanyam99ag:
That is a great recommendation to keep uniformity for better user experience.
So we can decrease the size of the calendar icon, so as to display the date in standard format i.e. YYYY/MM/DD

![Screenshot from 2020-05-03 00-29-05](https://user-images.githubusercontent.com/39384430/80873301-2979eb00-8cd5-11ea-9f8a-ddefdc27e2b5.png)

cgorringe:
I'm not in favor of reducing the size of the calendar icon. If we're going to do this, I'd rather widen the buttons or shrink the font size of the date.

I'd suggest the date format be whatever the default locale is for the user. Example:
```
[...]

PR #664, opened 2020-08-27
Adds datetime tooltips to Oldest and Newest buttons

cgorringe:
Implements #475 by adding datetime tooltips to the Oldest and Newest buttons.
Works only when Wayback Count is turned on.

### DIFF
--- a/webextension/manifest.json
+++ b/webextension/manifest.json
@@ -1,7 +1,7 @@
 {
   "manifest_version": 2,
   "name": "Wayback Machine",
-  "version": "3.0.0.12",
+  "version": "3.0.0.13",
   "description": "The Official Wayback Machine Extension - by the Internet Archive.",
   "icons": {
     "16": "images/app-icon/mini-icon16.png",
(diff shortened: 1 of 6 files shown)


tikhsuP (commented):
It works well!
I noticed that you also made the change that when on pages that have never been saved, the SPN button doesn't flip.

I have the following suggestions:
- We should update the tooltip timestamps when using the Search URL.
- We should update the tooltip timestamps after a successful save.
- Should we not flip the SPN button when on excluded URLs?

cgorringe:
OK, I implemented the following suggestions:
1. Cleared tooltips during Search.
2. Updated timestamps after Save.
3. Didn't change flipping behavior. Keep flip on Excluded URLs for now.
Also did these:
4. Display wayback count when setting checked on.
5. Now showing `0` in badge if valid page that hasn't yet been archived.

tikhsuP (approved):
Looks good!
Thanks for the changes.

PR #666, opened 2020-08-29
Unit test fixed

anishsarangi:
In this PR, I have added a new package `sinon-chrome` to fix "TypeError : Chrome is not defined issue". 
I have also fixed unit test errors in `annotation.js` and `tagcloud.js`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,70 +30,44 @@
       "integrity": "sha512-s5PLdI9NYgjBvfrv6rhirPHlAHWx+Sfo/IjsAeiXYfmemC/GSjwsyz1wLnGPazbLPXWfk62ks980o9AmsxYUEQ==",
       "dev": true
     },
-    "@internetarchive/radial-tree": {
-      "version": "1.0.0",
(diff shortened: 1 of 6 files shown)


cgorringe:
`sinon-chrome` may be a useful package, but afraid it may cause some issues as I'll need to convert all `chrome` references to `browser` and likely need to use the [mozilla/webextension-polyfill](https://github.com/mozilla/webextension-polyfill) to get working under Safari 14. I noticed someone posted [this issue with the polyfill](https://github.com/acvetkov/sinon-chrome/issues/103) and [these other issues](https://github.com/acvetkov/sinon-chrome/issues) so it may be a while before I can approve this, as I need to make the necessary changes and test under Safari 14.

nickmccurdy:
This fixed it for me https://github.com/mozilla/webextension-polyfill/issues/218#issuecomment-584936358

cgorringe:
I think the right solution would be just to not reference `chrome` directly in **utils.js** and do it from **background.js** instead of pulling in an external dependency.

anishsarangi:
Thank you @cgorringe . I am looking into it.

cgorringe:
Fixed this in PR #679 without needing to import `sinon-chrome`.

PR #692, opened 2020-12-05
Reduce available api calls frequency and use cached value

anishsarangi:
This PR is a fix to #474.


cgorringe (approved):
It looks great! I just did an extra check and made sure that the URL used matched what the wayback count badge used. Ready to Squash and Merge.

PR #693, opened 2020-12-13
More caching even if wayback count option is not checked

anishsarangi:
This PR is an improvement to #692 to introduce more caching


cgorringe (approved):
It's nice when finding a clean solution that reduces code! I hadn't run it yet, but reading through the code, it looks good.

PR #695, opened 2020-12-30
User-agent now changes dynamically based on the browser

anishsarangi:


cgorringe:
I'm fine with this change, but would like Mark's input on this. There may be a couple of other things to change as well.

To summarize: All versions of this extension currently add the string `Wayback_Machine_Chrome/` + _VERSION_ to the `User-Agent` http header, including browsers other than Chrome. This PR changes this to `Wayback_Machine_X/` where X is `Chrome` `Chromium` `Firefox` `Safari` `Edge` `Opera` or `Brave`.

This currently only occurs when calling `https://X-api.archive.org/wayback/available` where X is the browser. Since all API calls can already be distinguished by what browser is being used based on the subdomain, e.g. `chrome-api.archive.org`, it's not necessary to change the User-Agent string. Also note that we're currently only modifying the User-Agent string on just that 1 API call and not the others, while ALL the API calls currently use the browser-specific subdomains.

The native Safari (pre-14) Extension (which will be retired) currently uses these http headers:
```
User-Agent: Wayback_Machine_Safari_XC/APP_VERSION
Wayback-Extension-Version: Wayback_Machine_Safari_XC/APP_VERSION
Wayback-Api-Version: 2
```

So some questions that I have for what we should do:
1. Should we modify every API call header's User-Agent string? Currently using only the browser subdomain doesn't provide the extension's version number on the other API calls.
2. Should we add the 2 additional headers used in the native Safari Extension?

anishsarangi:
1. For now, we are only tracking for one API. I am modifying the User-agent because there are entries for brave-api and firefox-api that use the Wayback_Machine_Chrome user-agent. This is a mismatch.
2. If using http://archive.org... then you need the Wayback-Api-Version: 2  when using POST. For safari-api.archive.org and other {browser}-api.archive.org it doesn't matter.

cgorringe (approved):
OK, sounds good.
You can replace `getBrowser()` with `gBrowser` to avoid rerunning the function every time.
Approved.

Issue #690, opened 2020-12-04
Bulk save feature breaks because extension logs out


Melonadev:
Sometimes when I'm saving multiple pages using the 'Bulk save' feature, at some point it stops working due to it automatically **logging out of Internet Archive**.

![image](https://user-images.githubusercontent.com/66422959/101171863-59c6a500-367b-11eb-9dbf-1bee60285c62.png)

Why does the extension log my account out every 15-30 minutes? Is this because Microsoft Edge is suspicious of the developer-unpacked extension or is it a security feature?

cgorringe:
Thanks for catching this bug! I haven't encountered it before, likely because I don't have Edge on my Mac so it gets tested the least on that browser. The code isn't written to auto-logout, so I'm guessing it's something specific to Edge that's causing this.

Another possibility is if you were on the archive.org website and happened to log out there while the Bulk Save was going on, since it uses the same cookie which would get cleared if you logged out there.

Obviously I need to install Edge on my Mac to debug this...

Melonadev:
I'm logged into Internet Archive website on Edge all the time, and I checked that I was still logged into it when this issue occurred.

But every time I start Microsoft Edge, a message pops up at the extension bar, "Disable developer mode extensions", pointing to this extension. Does this occur in other browsers as well?

andrizmitnick:
btw this issue also happens on chrome canary

krackers:
I'm seeing this as well. I think some of the warnings are spurious because the extension shows the "looks like you aren't logged in" on any error, even if the underlying error isn't due to that. For instance, when calling the /status endpoint if IA return with "error:job-failed" then that will propagate up and cause the message to be displayed.

cgorringe:
Thanks everyone for testing and reporting this bug! Looks like we may need to better check for errors and pause the bulk saving so they don't cascade.

anishsarangi:
Thanks everyone for reporting this bug. I will look into this. Thank you !

cgorringe:
This is likely fixed now, but could be tested to make sure.

contextnerror:
I haven't been trying to use bulk save, but I've also been noticing the extension seems to be logging me out for no reason after a certain amount of time. I don't know exactly how long since I only ever notice when I open it up and save page now is suddenly greyed out. This is happening in the latest version of google chrome.

Melonadev:
The latest version (currently) is fully functional, and it doesn't even ask me to log in at all.

[...]

PR #700, opened 2021-01-16
Show login error when the error is only with login in Bulk Save

anishsarangi:
This is a fix to #690


cgorringe (approved):
I wish the API would give some kind of error code so we wouldn't have to rely on that specific error text. But it is what it is.

Seems to work OK, so approving.

PR #705, opened 2021-02-12
Added Media Icons for Browsers

dwaipayan05:


cgorringe:
Thanks @dwaipayan05 Good idea! I ended up using a different set of icons that were already in the project. Copied and merged from PR #712

PR #706, opened 2021-02-12
Added error messages for Fact Checks and Alexa Internet 

### LINKED ISSUE #703: Specific error message for fact checks
When clicking on fact check button if some network error occurs it still says:
`No fact checks found for the current URL.`
which is inappropriate.
We could add a  different error message for when such problem occurs

the-pro:
Resolves #703 

Gave a specific message for network errors
A page with no fact checks found still shows the original message


the-pro:
The Latest Commit Resolves #708 
The Alexa Internet  now displays error messages for network problems which was showing loading status earlier.
Here is a screenshot : 

Alexa_solve

@cgorringe Could you please review this PR

PR #707, opened 2021-02-20
Added New feature to show error message while doing bulk save

Yashraj1406:
Added a new feature which will show a error message while doing a bulk save->import all bookmarks if the browser doesn't contain any bookmarks

### DIFF
--- a/webextension/scripts/bulk-save.js
+++ b/webextension/scripts/bulk-save.js
@@ -87,6 +87,7 @@ function processTreeNode(node) {
   }
   // add bookmark URL from leaf node
   if (node.url && isValidUrl(node.url) && isNotExcludedUrl(node.url) && !isDuplicateURL(node.url)) {
+    count++
     addToBulkSave(node.url)
   }
 }
--- a/webextension/scripts/bulk-save.js
+++ b/webextension/scripts/bulk-save.js
@@ -70,14 +70,14 @@ function importAllBookmarks() {
     chrome.bookmarks.getTree((nodeTree) => {
       nodeTree.forEach((node) => {
         processTreeNode(node)
-        //Check and response for no bookmark in the browser
+        // Check and response for no bookmark in the browser.
         if (Object.keys(bulkSaveObj).length === 0) {
-          alert("Sorry No Bookmarks Found. Please ensure that you have bookmarks");
+          alert('Sorry No Bookmarks Found. Please ensure that you have bookmarks')
         }
       })
     })
   }
-} 
+}
 // Traverses the bookmark tree nodes recursively.
 function processTreeNode(node) {
   // process child nodes
--- a/webextension/scripts/bulk-save.js
+++ b/webextension/scripts/bulk-save.js
@@ -70,6 +70,9 @@ function importAllBookmarks() {
     chrome.bookmarks.getTree((nodeTree) => {
       nodeTree.forEach((node) => {
         processTreeNode(node)
+        if (count === 0) {
+          alert("Sorry No Bookmarks Found. Please ensure that you have bookmarks")
+        }
       })
     })
   }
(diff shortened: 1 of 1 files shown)


Yashraj1406:
hello there, @cgorringe please review this and let me know about it

anishsarangi:
Hello @Yashraj1406 . Thank you for your PR. There is still a bug you need to resolve. 
When clicked on "Import all bookmarks", the alert is coming only for the first time, not everytime. Please have a look at it !

Yashraj1406:
Thank you @anishsarangi for reviewing it, please look into the changes I have made, I hope it is working fine now

anishsarangi:
@Yashraj1406 when I cleared all the bookmarks, and then again tried to "Import all bookmarks", I am not seeing the alert message. `count` is not getting reset.

Yashraj1406:
@anishsarangi I have modified it, please review this

webextension/scripts/bulk-save.js
```diff
@@ -70,11 +70,16 @@ function importAllBookmarks() {
     chrome.bookmarks.getTree((nodeTree) => {
       nodeTree.forEach((node) => {
         processTreeNode(node)
+        if (count === 0) {
```
anishsarangi, line 73: Instead of storing the count variable, you could check the length by `Object.keys(bulkSaveObj).length`. We already have a `bulkSaveObj` where we are storing the URLs

anishsarangi (changes requested):

webextension/scripts/bulk-save.js
Yashraj1406: @anishsarangi Thank you for guiding me

anishsarangi (approved):

webextension/scripts/bulk-save.js
```diff
@@ -70,11 +70,14 @@ function importAllBookmarks() {
     chrome.bookmarks.getTree((nodeTree) => {
       nodeTree.forEach((node) => {
         processTreeNode(node)
+        //Check and response for no bookmark in the browser
+        if (Object.keys(bulkSaveObj).length === 0) {
+          alert("Sorry No Bookmarks Found. Please ensure that you have bookmarks");
```
anishsarangi, line 75: Please run eslint before a commit to be consistent with the code styles. (e.g you have an extra `;` here)

Yashraj1406: Sure will do that, and Thank you for guiding me

cgorringe:
Using `Object.keys(bulkSaveObj).length` is incorrect, as `bulkSaveObj` is where the URLs from the text area are stored, which may include user-entered URLs that are not in the bookmarks. Using a count variable was correct, although think about not using a global count var but instead have the `processTreeNode()` function return the count number back to `importAllBookmarks()` which returns it to `doImportBookmarks()` which is where I recommend placing an alert.

anishsarangi:
Thank you @cgorringe . I missed the fact that the URLs from the text area are also stored in `bulkSaveObj`.

Yashraj1406:
@cgorringe I have created the global variable so that it is able to avoid the re-initialization while recursion of processTreeNode(), if there is anything in javascript which helps to avoid this re-initialization (something of sort static int in C) please let me know I will be very grateful

cgorringe:
@Yashraj1406 It's OK to create a local variable in the recursive function, as any decrease in speed or memory is small and not noticeable.

I *think* this may be the correct code, but please double-check...
```
function processTreeNode(node) {
  let count = 0
  // process child nodes
  if (node.children) {
    node.children.forEach((child) => { count += processTreeNode(child) })
   }
   // add bookmark URL from leaf node
   if (node.url && isValidUrl(node.url) && isNotExcludedUrl(node.url) && !isDuplicateURL(node.url)) {
     count++
     addToBulkSave(node.url)
   }
   return count
 }
```

If value returned from first call to proccessTreeNode() == 0 then no bookmarks found.

Yashraj1406:
@cgorringe Thank you very much, I think the above changes work perfectly but I wanted to know whether we have to change the function ### importAllBookmarks() to ### async function because while returning the value from ### importAllBookmarks() to###  doImportBookmarks it is not waiting for function ( as it is a callback function ) to return a value , So please suggest me that do we have change the function ### importAllBookmarks() to ### async one and then pass the value by using await OR the above changes are fine.

cgorringe:
@Yashraj1406 Good point, I missed that bookmarks.getTree() returned a Promise. It's OK to keep the alert() in importAllBookmarks() and the code looks good to me as it is now!

Issue #691, opened 2020-12-05
Option to save further outlinks of each url in Bulk save


Melonadev:
The extension itself already supports archiving outlinks of a webpage the user is visiting. So why not also do that for each url in Bulk save?

Example: [Madness Serial](https://madnessserial.com) is a website that offers free e-books, some of which are editions of existing public domain books.

Take this page for [Moby Dick](https://madnessserial.com/mdash/moby-dick-herman-melville). Here're the download formats:
![image](https://user-images.githubusercontent.com/66422959/101246243-0886e600-374d-11eb-9598-bfd2d5f0f7ab.png)

The trouble is that the [PDF reader](https://static1.squarespace.com/static/5ea988cb2f98c73c672bd6bb/t/5fb05d9d9a1aef48f68024e4/1605393892092/Moby+Dick%3B+or%2C+The+Whale+-+Herman+Melville.pdf) is not hosted on Madness Serial, but on a Squarespace domain. That's the reason why the [XML sitemap generator that I use](https://www.web-site-map.com/) doesn't include it.

That PDF is in just one of the non-Madness-Serial outlinks from the earlier Moby Dick page, and many of them are not in my list of urls. This is where this feature I'm suggesting comes in.

cgorringe:
Thanks for the suggestion! Yes we can add the Outlinks and Screenshot options to Bulk Save. (although Outlinks will slow it down a bit.)

Melonadev:
Exploring this idea further, let's say that the url list is like this:

**url 1** (outlinks are **url 2** & **url 3**)
**url 2** (outlink is **url 3**)
**url 3** (outlinks are **url 1** & **url 4**)
**url 4**
**url 5**

How about adding a _time-saving option_ (maybe make this the default?) where, instead of saving every single outlink for each url without checking it's been saved already in the same session, the extension does this:

1. Archive the first url, which is **url 1**, and detect and archive its outlinks **url 2** and **url 3**.
2. Proceed to the next url in the list, which is **url 2**.
3. Detect that **url 2** has already been archived earlier _this session_.
4. _Skip_ archiving **url 2**, but continue to detect its outlink **url 3** (also skipped because already archived earlier this session).
5. Proceed to the next url in the list, which is **url 3**.
6. _Skip_ **url 3**, but continue to detect its outlinks **url 1** and **url 4**.
7. _Skip_ **url 1**, but archive **url 4**.
8. Proceed to the next url in the list, which is **url 4**.
[...]

PR #719, opened 2021-04-04
Added Outlink and Screenshot option to Bulksave

Yashraj1406:
This is issue number #691. Added Outlink and Screenshot option to Bulksave

### DIFF
--- a/webextension/bulk-save.html
+++ b/webextension/bulk-save.html
@@ -42,7 +42,18 @@
 
       <div class="save-box">
         <div class="save-options-container">
-          <p><input type="checkbox" id="never-saved"> <label class="save-options" for="never-saved">Save only URLs that have never been archived.</label></p>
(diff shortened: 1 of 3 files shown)


Yashraj1406:
@cgorringe, Please review this

cgorringe:
Testing. Getting rate limiting bugs. I'll try to fix.

Yashraj1406:
Thank you @cgorringe, but is there anything from my side which I can fix as I will be more than happy to learn and contribute

cgorringe:
Thanks @Yashraj1406, after trying it out, I think saving Outlinks isn't going to work right now because while the server is processing the outlinks, it hits the save limit when trying to save more URLs. Maybe we can process the outlinks one by one in the extension, but for now I think we should leave it out.

Saving the screenshot works, so I kept that in.

PR #710, opened 2021-02-26
Added Error Messages for Alexa Internet and Fact Checks

the-pro:
Resloves #703 

Added error messages to fact checks for network errors.

Resolves #708 
Added Error messages for Alexa Internet for network errors.

Alexa_solve

PR #725, opened 2021-04-09
Change the title when no fact checks available for an URL

anishsarangi:
I have removed the message "Cannot reach the server at the moment !" as the `error` object in the response means no fact checks found for a given URL.


cgorringe:
Linking issue #703 and PR #710.
Fixed it so that the correct error message should show now.

cgorringe (approved):

anishsarangi:
Thank you @cgorringe

Issue #729, opened 2021-04-12
Bulk Save doesn't archive links containing special characters


Melonadev:
Examples of links with entirely 'normal' Unicode characters in the English language:

- https://www.bravovail.org/tickets-performances/2018/orchestra-concerts/dallas-symphony-orchestra/
- https://www.bravovail.org/tickets-performances/2018/orchestra-concerts/dallas-symphony-orchestra/women-rock/
- https://www.bravovail.org/tickets-performances/2018/orchestra-concerts/dallas-symphony-orchestra/tchaikovsky-s-fifth-symphony/

Examples of links including 'special' Unicode characters outside the English language (**bold** is original, normal is converted):

- **https://www.bravovail.org/tickets-performances/2018/orchestra-concerts/dallas-symphony-orchestra/măcelaru-conducts-rachmaninoff/**
...is converted to:
https://www.bravovail.org/tickets-performances/2018/orchestra-concerts/dallas-symphony-orchestra/m%C4%83celaru-conducts-rachmaninoff/

- **https://www.bravovail.org/tickets-performances/2018/orchestra-concerts/the-philadelphia-orchestra/denève-conducts-brahms-dvořák/**
...is converted to:
https://www.bravovail.org/tickets-performances/2018/orchestra-concerts/the-philadelphia-orchestra/den%C3%A8ve-conducts-brahms-dvo%C5%99%C3%A1k/

- **https://www.bravovail.org/tickets-performances/2018/orchestra-concerts/the-philadelphia-orchestra/thibaudet-plays-saint-saëns/**
...is converted to:
https://www.bravovail.org/tickets-performances/2018/orchestra-concerts/the-philadelphia-orchestra/thibaudet-plays-saint-sa%C3%ABns/

And as you can see from the screenshot below, the Bulk Save function struggles to archive the latter three links, even long after the other links are done:

![image](https://user-images.githubusercontent.com/66422959/114370501-21e3e700-9bb2-11eb-94fc-5b300c1344f9.png)

cgorringe:
Thanks for catching this!

Upon investigation testing this on Chrome, I discovered that entering the URLs containing Unicode manually and tapping "Add URLs" works OK. But when I loaded one of them in the browser, save as a bookmark, then "Import All Bookmarks", the URL added to the list is %-encoded. Opening Chrome's Bookmark Manager, I found that Chrome stores the URL as %-encoded! So clearly Chrome's fault!

[...]

PR #730, opened 2021-04-13
Bulk Save fix to import bookmarks containing encoded URLs

cgorringe:
Fix for #729.


cgorringe:
Fixed savePageNow() to work correctly with URL encoding.

anishsarangi (approved):
Looks good and working fine.

PR #738, opened 2021-04-20
Settings tab tooltip to display on first popup

cgorringe:
After loading (or reloading) the extension for the first time, after agreement, the popup will display a tooltip over the Settings tab icon in order to direct the user's attention to Settings. After hovering over the Settings tab icon, the tooltip will be turned off on all future popups.


anishsarangi (approved):
Nice addition.

PR #742, opened 2021-04-26
Show Cited Papers button when there are papers but No Books.

### LINKED ISSUE #718: Show Cited Papers button when there are papers but No Books
**Describe the bug**

While testing Wikipedia Resources, if there are No Books (e.g. ISBN IDs) on the page, but there are Papers (e.g. "doi:" IDs) on a Wikipedia page, the 'R' icon will not show, and as a result the 'Cited Papers' button will not display.

**To Reproduce**

Example page where there should be Cited Papers, but No Books.
https://en.wikipedia.org/wiki/Perseverance_(rover)

**Expected behavior**

One possible solution: Scan for ISBN IDs for Books, and DOI (and others) for Papers, without calling an API, as an initial step perhaps before calling the API. The 'Cited Books' and 'Cited Papers' buttons currently either both display or not at the same time, but it's not always the case that both would be present, so we may want to only display one or the other. If we scan for the IDs before calling the API, this would lead to displaying the 'R' icon and Cited buttons more quickly.

anishsarangi:
Fix #718


cgorringe:
The responsiveness of the blue buttons showing up is too slow. Needs to display at same time as the rest of the UI when popup displays.

cgorringe:
Doesn't do the extra step of "Scan for ISBN IDs for Books, and DOI (and others) for Papers, without calling an API", but it does fix the primary issue now.

PR #750, opened 2021-06-03
Fix false 404 popup

### LINKED ISSUE #748: False 404 popup
**Describe the bug**
Sometimes, the 404 popup is showing on a non-404 URL. 

Screenshot 2021-05-17 at 11 55 50 PM

**To Reproduce**
1. Go to https://www.creativebloq.com/web-design/best-404-pages-812505
2. Click on any sample 404 URL on the page (click and open link in a new tab)
3. User will see the 404 banner on the current window (not in the other tab where it should show)

**Expected behavior**
User should see the 404 banner on the correct tab.

anishsarangi:
Fixed #748

### DIFF
--- a/webextension/scripts/background.js
+++ b/webextension/scripts/background.js
@@ -402,7 +402,7 @@ chrome.webRequest.onCompleted.addListener((details) => {
       }, () => {})
     }
   }
-  if(details.tabId > 0){
+  if (details.tabId > 0) {
     chrome.storage.local.get(['not_found_setting', 'agreement'], (settings) => {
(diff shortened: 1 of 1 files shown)


cgorringe:
I'm getting the same error as this person in this post, which looks like may be a bug in Chrome 91 (most recent version).
https://stackoverflow.com/questions/67806779/im-getting-an-error-tabs-cannot-be-edited-right-now-user-may-be-dragging-a-ta

cgorringe (approved):
Tested under Chrome and Firefox, and works fine (besides that one issue with Chrome, which may be unrelated.)

PR #751, opened 2021-06-07
added missing utf8 encoding declaration

cgorringe:
This fixes the following Firefox console error when **dnserror.html** is loaded when a domain doesn't exist.

> The character encoding of the HTML document was not declared. The document will render with garbled text in some browser configurations if the document contains characters from outside the US-ASCII range. The character encoding of the page must be declared in the document or in the transfer protocol.


anishsarangi (approved):
Looks good

Issue #769, opened 2021-09-09
Safari 14 bugs


cgorringe:
Bugs remaining in the Safari 14 port:

- [x] SPN flip button backside text visible.
- [x] Popup buttons and Settings hover glitches.
- [x] Replace 404s, etc broken.
- [x] Clicking links in "Show Resources During Save" and Bulk Save windows opens link in the same window instead of main browser window.
- [x] Feedback link in popup points to iOS app on App Store instead of Safari extension.

PR #771, opened 2021-09-16
Variety of Safari bug fixes

cgorringe:
Strangely, there's some glitching and slowdowns while hovering over popup buttons in Chrome now, where before this only occurred in Safari. Likely due to the massive `style.css` file which I'll work on next.

There's a bug in Safari regarding `manifest.js`: Under `"content_scripts"` the `"matches"` field appears to be ignored by Safari and js and css files in this section are inserted in ALL websites instead of just Wikipedia. Seems to only occur when "Always Allow on Every Website..." button is selected in preferences (which is recommended) but this causes all sorts of problems on websites...

### DIFF
--- a/webextension/css/index.css
+++ b/webextension/css/index.css
@@ -259,12 +259,11 @@ a.btn-social, .btn-social {
 }
 
 .btn-social .fa, .btn-social i {
-    backface-visibility: hidden;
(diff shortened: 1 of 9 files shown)


cgorringe:
fixes #769

PR #782, opened 2021-09-27
Remove checking the current active tab-url and tabId in amazon book feature

### LINKED ISSUE #779: Amazon books feature not working all the time
**Describe the bug**
I noticed an issue with the amazon books feature. It is not working all the time.
In the background.js network tab, i could see that the api is not called. But when i reload the URL, it is working and calling the api.

**To Reproduce**
1. Turn on the amazon books feature in the extension.
2. Go to [testing guide URL](https://github.com/internetarchive/wayback-machine-webextension/blob/master/TESTING_GUIDE.md#manual-tests) and then go to Amazon Books section.
3. Click on(click and open in a new tab) all the urls
4. Then go to each URL to check if there is “R” icon and a button called “Read Book”
**Expected behavior**
There should be a "R" icon and "Read Book" button in the extension

**Screenshots**
Screenshot 2021-09-25 at 12 00 31 AM
Screenshot 2021-09-25 at 12 00 38 AM
**Desktop**
 - OS: [Mac]
 - Browser [chrome]
 - Version [93.0.4577.82]

anishsarangi:
Fix #779

### DIFF
--- a/webextension/scripts/background.js
+++ b/webextension/scripts/background.js
@@ -616,25 +616,21 @@ chrome.tabs.onUpdated.addListener((tabId, info, tab) => {
       if (open_url.slice(-1) === '/') { open_url = received_url.substring(0, open_url.length - 1) }
       chrome.storage.local.get(['amazon_setting'], (settings) => {
         // checking amazon books settings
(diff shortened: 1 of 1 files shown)


cgorringe (approved):
looks good

PR #785, opened 2021-09-29
Fix Unchecked runtime.lasterror console errors 

anishsarangi:


webextension/scripts/utils.js
```diff
@@ -443,19 +443,27 @@ function opener(url, option, callback) {
...
     type: 'basic',
     title: 'WayBack Machine',
     message: message,
     iconUrl: chrome.runtime.getURL('images/app-icon/app-icon96.png')
   }
-  chrome.notifications && chrome.notifications.create(options, callback)
+  chrome.notifications && chrome.notifications.create(options, callbackFn)
+}
+
+function callback() {
```
infojunkie, line 462: Let's call this function something more descriptive, such as `errorLog`. In this case, no need to rename existing functions

webextension/scripts/resource_list.js
```diff
@@ -24,6 +24,7 @@ function showResourceData(url_name) {
   $('#current-url').text(url_name)
   chrome.runtime.onMessage.addListener(
     (message) => {
+      if (chrome.runtime.lastError) {}
```
infojunkie, line 27: Are you sure we don't want to log this?

webextension/scripts/utils.js
```diff
@@ -443,19 +443,27 @@ function opener(url, option, callback) {
...
+  if (chrome.runtime.lastError) {
+      console.log(chrome.runtime.lastError.message);
```
infojunkie, line 464: `console.error` sounds more correct here

webextension/scripts/utils.js
```diff
@@ -443,19 +443,27 @@ function opener(url, option, callback) {
...
+  } else {
+      // Tab exists
```
infojunkie, line 466: This callback is not only about the tab existing or not, correct? It applies to _any_ `lastError`.

webextension/scripts/background.js
```diff
@@ -344,7 +344,7 @@ function getCachedFactCheck(url, onSuccess, onFail) {
 chrome.runtime.onStartup.addListener((details) => {
   chrome.storage.local.get({ agreement: false }, (settings) => {
     if (settings && settings.agreement) {
-      chrome.browserAction.setPopup({ popup: chrome.runtime.getURL('index.html') })
+      chrome.browserAction.setPopup({ popup: chrome.runtime.getURL('index.html') },callback)
```
infojunkie, line 347: Would be great to leave a space after the comma, as per the remainder of this codebase.

webextension/scripts/archive.js
```diff
@@ -85,6 +85,7 @@ function refreshWayback(url, code) {
 // Listens to SHOW_BANNER messages
 chrome.runtime.onMessage.addListener(
   (request, sender, sendResponse) => {
+    if (chrome.runtime.lastError) {}
```
infojunkie, line 88: Are you sure we don't want to log this?

webextension/scripts/background.js
```diff
@@ -391,6 +391,7 @@ chrome.webRequest.onErrorOccurred.addListener((details) => {
 
 // Listens for website loading completed for 404-Not-Found popups.
 chrome.webRequest.onCompleted.addListener((details) => {
+  if (chrome.runtime.lastError) {}
```
infojunkie, line 394: Are you sure we don't want to log this?

webextension/scripts/background.js
```diff
@@ -439,6 +440,7 @@ chrome.webRequest.onCompleted.addListener((details) => {
 // note: return true only if sendResponse() needs to be kept around.
 //
 chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
+  if (chrome.runtime.lastError) { console.log("capturing the error") }
```
infojunkie, line 443: Why is logged differently?

webextension/scripts/background.js
```diff
@@ -569,6 +571,7 @@ chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
 })
 
 chrome.tabs.onUpdated.addListener((tabId, info, tab) => {
+  if (chrome.runtime.lastError) { }
```
infojunkie, line 574: Are you sure we don't want to log this?

webextension/scripts/bulk-save.js
```diff
@@ -222,6 +222,7 @@ function isDuplicateURL(url) {
 // Listens to SPN response messages from background.js.
 function initMessageListener() {
   chrome.runtime.onMessage.addListener((msg) => {
+    if (chrome.runtime.lastError) { }
```
infojunkie, line 225: Are you sure we don't want to log this?

webextension/scripts/popup.js
```diff
@@ -643,6 +629,7 @@ function setupViewSetting() {
 function setupSaveListener() {
   chrome.runtime.onMessage.addListener(
     (message) => {
+      if (chrome.runtime.lastError) { /* skip */ }
```
infojunkie, line 632: Are you sure we don't want to log this?

webextension/scripts/resource_list.js
infojunkie, line 27: Are you sure we don't want to log this?

webextension/scripts/utils.js
anishsarangi: Renamed it to `checkLastError`

webextension/scripts/utils.js
cgorringe: I'd keep it at `console.log` since many of these will be "false alarms" which we were trying to suppress in the first place.

webextension/scripts/utils.js
```diff
@@ -458,6 +459,14 @@ function notify(message, callback) {
   chrome.notifications && chrome.notifications.create(options, callback)
 }
 
+function checkLastError() {
+  if (chrome.runtime.lastError && chrome.runtime.lastError.message.startsWith("No tab with id:")) {
```
infojunkie, line 463: I don't understand the intent of this line: why do we want only log the `lastError` if it's a tab error, but ignore it in case of other errors?

anishsarangi: I have updated the function in the latest commit. Please check this out. Thanks

webextension/scripts/utils.js
```diff
@@ -460,8 +460,12 @@ function notify(message, callback) {
 }
 
 function checkLastError() {
-  if (chrome.runtime.lastError && chrome.runtime.lastError.message.startsWith("No tab with id:")) {
+  if (chrome.runtime.lastError) {
+    if (chrome.runtime.lastError.message.startsWith("No tab with id:")){
       console.log(chrome.runtime.lastError.message);
```
infojunkie, line 465: That's fine with me, although I would have thought we want to _ignore_ such errors (i.e. not log them)

webextension/scripts/utils.js
infojunkie: I think this comment is inaccurate. The correct comment would read:
`// No error occurred`

webextension/scripts/utils.js
anishsarangi: Yes, we can ignore them. I will remove the console. statement. I added it for testing purpose.

cgorringe (approved):

cgorringe:
Note for future reference:
- The downside of calling a function is that we don’t get the line number of the error in console log.
- Can get by calling console.trace() inside the function. 
- Changed console.error to console.log because the point is to not alert the user and needed to pass review.

PR #786, opened 2021-09-30
Add 404 setting options to private_before_default and remove others

anishsarangi:
If someone reload the extension OR install it new, by default the “404 option is checked” in the settings. Then if user is selecting some more options and then if the user is selecting and deselecting private mode, you can see that all other checkboxes are checked. It is not restoring to its previous state.

This PR is a fix to that


cgorringe (approved):
looks fine

PR #788, opened 2021-10-04
Show host instead of url in tagcloud page

anishsarangi:


webextension/scripts/tagcloud.js
```diff
@@ -17,7 +17,7 @@ function getTags() {
   let y = hostname.split('.')
   let not_display4 = y.join(' ')
   let not_display1 = y.join(' ')
-  $('.url').text(url).attr('href', url)
+  $('.url').text(hostname).attr('href', hostname)
```
cgorringe, line 20: If you click on the link, it fails due to missing `http(s)://`. Could keep the href pointed to `url`

webextension/tagcloud.html
```diff
@@ -15,7 +15,7 @@
       <h2 id="main-title">Word Cloud</h2>
       <p id="main-info"></p>
       <p id="error-msg"></p>
-      <h4>URL: <a class="url" target="_blank"></a></h4>
+      <h4>HOST: <a class="url" target="_blank"></a></h4>
```
cgorringe, line 18: could just remove `HOST:` label

cgorringe (approved):

PR #789, opened 2021-10-05
Remove Bulk Save and rearrange other popup buttons

cgorringe:
Also tweaked a label in the Bulk Save window due to user feedback.

### DIFF
--- a/webextension/scripts/background.js
+++ b/webextension/scripts/background.js
@@ -312,6 +312,7 @@ function getCachedTvNews(url, onSuccess, onFail) {
 }
 
 function getCachedFactCheck(url, onSuccess, onFail) {
+  // require "https://data.our.news/api/*" in manifest permissions
   const requestUrl = 'https://data.our.news/api/?partner=wayback&factcheck=' + fixedEncodeURIComponent(url)
   fetchCachedAPI(requestUrl, onSuccess, onFail)
 }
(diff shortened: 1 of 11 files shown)


anishsarangi:
We can also remove the fact-check option from the settings. We can also remove it from the permissions in manifest.json

webextension/scripts/background.js
```diff
@@ -312,6 +312,7 @@ function getCachedTvNews(url, onSuccess, onFail) {
 }
 
 function getCachedFactCheck(url, onSuccess, onFail) {
+  // require "https://data.our.news/api/*" in manifest permissions
```
infojunkie, line 315: We won't be using this service going forward, even when we reinstate the fact-checking feature. So this comment is no longer relevant.

cgorringe:
updated code ready for review

webextension/scripts/overview.js
```diff
@@ -27,10 +26,9 @@ function getArchiveDetails (url) {
     .text(captures.toLocaleString())
     if (captures > 0) {
       $('#total_captures').show()
-      getThumbnail(url)
+      // getThumbnail(url)
```
anishsarangi, line 29: Are we not going to show the thumbnail in overview context ?

anishsarangi:
We can remove the reference of our.news from the about page.

webextension/scripts/overview.js
cgorringe: I commented it out because it's not over a secure connection and remember reading about api calls should be over https or might not pass review. (could be wrong...)

cgorringe: But you know I'm probably just being overly cautious and would like to see thumbnails shown, so I will restore it back and if it passes then great.

PR #793, opened 2021-10-06
Remove extra callbacks when there is no response and fixed a minor bug in settings.js

anishsarangi:


cgorringe (approved):
Tested in Firefox. Looks good! Approved for merge.

Issue #774, opened 2021-09-21
Wikipedia popup text extends outside border


infojunkie:
**Describe the bug**
The text within the popup that's shown upon hovering over the green/black IA icon on Wikipedia, extends outside the boundaries of the popup.

**To Reproduce**
1. Enable the "Wikipedia Resources" option
2. Open https://en.wikipedia.org/wiki/World_War_II and scroll down to References
3. Hover over any green/black IA icon that's injected by the extension
4. Observe that the text within the popup appears cut off because it extends outside the popup borders

**Expected behavior**
The text should not be cut off nor extend outside the popup borders.

**Screenshots**
![Screenshot from 2021-09-21 14-28-26](https://user-images.githubusercontent.com/92977/134255152-592569b7-b926-432d-8b97-4188bddb2f2c.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - Browser: Chrome 93.0.4577.82

cgorringe:
Thanks. I noticed that too but forgot to report it!

cgorringe:
fixed #794

Issue #756, opened 2021-08-02
The add-on causes Wikipedia to resize on Firefox


unbeatable-101:
**Describe the bug**
If you have the add-on enabled, it causes Wikipedia to resize after you load the site

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to https://en.wikipedia.org
2. Look for the page resizing
3. Disable the add-on from `about:addons`
4. Reload the page again and see it doesn't resize

**Expected behaviour**
The page doesn't resize

**Videos**

https://user-images.githubusercontent.com/58151048/127813534-b4dda7ad-43a7-46ee-8578-a0fba10e3a3f.mov

**Desktop (please complete the following information):**
 - OS: macOS 12.0 Beta (21A5294g)
 - Browser: Firefox Nightly
 - Version: 92.0a1 (2021-08-01) (64-bit)

cgorringe:
Bug tested and verified to occur on Wikipedia using latest webextension (3.0) on both Chrome and Firefox.

On Firefox: Just like in the video, clicking Reload would increase the font size on the page briefly, then immediately go back to default font size. Doesn't do this when the extension is unloaded.

On Chrome: Shrinks the page's font size permanently when on. Turn off the extension, goes back to larger font size.

Thanks for the bug report!

cgorringe:
Figured out why this occurs is due to **style.css** being inserted into Wikipedia pages due to `manifest.json`. Fix would be to only include minimum necessary for the icon popups that are embedded in the page. **style.css** is a massive file. Don't need all that.

cgorringe:
Related to this issue: 

There's a bug in Safari regarding manifest.js: Under "content_scripts" the "matches" field appears to be ignored by Safari and js and css files in this section are inserted in ALL websites instead of just Wikipedia. Seems to only occur when "Always Allow on Every Website..." button is selected in preferences (which is recommended) but this causes all sorts of problems on websites.

Above should be fixed once this issue is fixed.

cgorringe:
Bug specific to Safari above may have been fixed in Safari 15.0.

cgorringe:
fixed #794

PR #794, opened 2021-10-07
Wikipedia book popup rewrite to improve efficiency

cgorringe:
Rewrote code that inserts icons and book popups on Wikipedia pages. Removes all dependencies, no longer freezes during rendering of long book lists, fixes font resizing bug #756, fixes popup text extending outside border #774, tweak styling, removed green donate popups and only display books found. Currently limits to work on only English Wikipedia.


cgorringe:
I know there's some glitches still with the popup sometimes being clipped or height being stretched. So I guess this isn't ready yet as I'm still trying to fix the UI.

PR #814, opened 2021-10-18
Removed 'tabs' permission from manifest

cgorringe:
Everything seems to work fine with 'tabs' permission removed from manifest, since the 'activeTab' permission allows use of tabs. Could use further testing in all 3 browsers.


anishsarangi (approved):
Working fine in both chrome and firefox. I didn't notice any issues.

cgorringe:
Tested and works under Safari

PR #828, opened 2021-10-25
Support ISBN parsing on multi-language Wikipedia sites

### LINKED ISSUE #791: Support all *.wikipedia.org sites for book parsing
Currently the extension searches for the anchor tags with title attribute containing `Special:BooksSources` in the Wikipedia sites to add the tooltip in the UI. But the webpages which are in other languages e.g (https://fr.wikipedia.org/wiki/Rongorongo), it doesn't work.

Currently it only works for English Wikipedia, need to support other languages too.

cgorringe:
fixes #791 in the web extension.

The server API needs to be updated to support these still, so until that is done, this code will not return any Cited Books nor draw any popup icons just yet.

Currently supports these Wikipedia sites:
English (en), Simple English (simple), Arabic (ar), German (de), Spanish (es), French (fr), Italian (it), Japanese (ja), Polish (pl), Portuguese (pt), Russian (ru), Chinese (zh).

There are a LOT of language sites to cover. More may be added in the future.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,6 +12,7 @@
     "prefer-const": "off",
     "quote-props": "off",
     "one-var": "off",
+    "padded-blocks": "off",
     "dot-notation": "off"
   },
   "globals": {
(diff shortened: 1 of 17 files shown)


cgorringe:
OK. I removed the language-specific matching strings and came up with a more generic solution that (so far as I can tell) works across all language sites.

Code also updated to use new POST API call, which will work once server is updated.

PR #832, opened 2021-11-30
Replace 404 embed popups

cgorringe:
This replaces the 404 embedded popup with a new toolbar 'V' icon and button "View Archived Version" within the toolbar popup. This is more consistent with the design of the rest of the extension.

It eliminates the possibility for the html embedded popup to interfere with web pages, which was a common complaint with the old deployed versions. Html code injection is removed on all but Wikipedia sites.

Known Issues:
- Firefox may require reloading a 4xx page for 'V' to display.
- Safari will not work with 'Server Not Found' errors. (and didn't previously)

404-not-found-screenshot


infojunkie:
Worked for me, thanks. My only comment is to question the choice of "V" to signify "View archived version". In my opinion, this letter has no explanatory power related to the signal we want to send to the user.

I would suggest to consider showing icons, instead of letters, for the vairous signals we want to send. E.g. in this case, adapting a [typical 404 icon](https://www.google.com/search?q=404+icon)

Issue #673, opened 2020-10-04
Auto-Update Previously Saved Pages


weirdalsuperfan:
**Is your feature request related to a problem/ issue? Please describe.**
So a cryptocurrency I (thankfully) didn't invest in recently exit-scammed, but upon checking their website, which I believe I had visited on my computer, only an old version from before their actual website was up was saved by the wayback machine, so there's little record of them beyond Google's image and search results cache, etc. All that the wayback machine had was like this domain name search result webpage or something.

Basically, websites can change, so just because they're saved doesn't mean they've REALLY been saved.

**Describe the solution you'd like**
Since there's no easy way for you to know if a website has been properly saved (you'd have to check all the snapshots manually), I'd like an auto-update option to save webpages even if they already exist.

**Describe alternatives you've considered**
I've considering making the user manually save the page, but that kind of defeats the purpose of this extension, so...yeah.

**Additional context**
The auto-update could consider: when the last snapshot was taken (more than a month ago? maybe it's changed), how many snapshots there have been (only one? might want to save to be safe), whether the page has changed since the last save (not sure if you can check this...), whether the page has ever been saved by you/your browser before (no? might want to save it just in case)

cgorringe:
Since we DO have an auto-save feature, but right now it only saves if the page had never been saved before. But since the check for this also returns a timestamp when it was last saved by the Wayback Machine, it wouldn't be that hard to implement. It's just a matter of adding a field to the Settings, or maybe a dropdown box to choose how long ago since the last snapshot. (e.g. "1 month", "1 year")

Thanks for the suggestion!

PR #833, opened 2021-12-02
Auto Save Age dropdown in Settings

cgorringe:
Implements #673 

This enhances the Auto Save Page feature by adding a dropdown list to Settings. Now instead of only auto-saving if a website has never been archived before, the user can specify an age such as 1 year, 90 days, or as soon as 24 hours. This would activate Auto Save if the latest archive is at least that age or older.

auto-save-age-screenshot

Known Issue: Setting Private Mode may not correctly clear Auto Save. This will be fixed in a future PR.

### DIFF
--- a/webextension/scripts/settings.js
+++ b/webextension/scripts/settings.js
@@ -16,10 +16,10 @@ $(function() {
   $('#view-setting').children('input,label').click((e) => { e.stopPropagation() })
   $('input[type=checkbox],input[type=radio]').change(saveOptions)
   $('#auto-archive-age').change((e) => {
-    $('#auto-archive-setting').prop('checked', true)
+    $('#auto-archive-setting').prop('checked', true).addClass('selected-prior')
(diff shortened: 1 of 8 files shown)


webextension/scripts/settings.js
```diff
@@ -15,6 +15,12 @@ $(function() {
   $('#view-setting').click(switchTabWindow)
   $('#view-setting').children('input,label').click((e) => { e.stopPropagation() })
   $('input[type=checkbox],input[type=radio]').change(saveOptions)
+  $('#auto-archive-age').change((e) => {
+    $('#auto-archive-setting').prop('checked', true)
```
anishsarangi, line 19: $('#auto-archive-setting').addClass('selected-prior')
This addition would fix the issue of private mode not clearing check-box.

cgorringe:
Tested in Chrome, Firefox, and Safari 15.

PR #839, opened 2021-12-09
Add 522 to the error code dict

anishsarangi:
For some urls like https://www.projet-voltaire.fr/regles-orthographe/va-ou-vas/ gives the archived version notice but shows "undefined" at the top of the notice. It is because 522 status code is not there in the `ERROR_CODE_DIC` in the code.
This is a fix for it.


infojunkie:
Thanks. Please make sure that all HTTP satus codes are covered, as per https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

cgorringe:
Since this will likely be replaced with PR #832, I'll update that code. No need to spend time adding more status codes, I got it.

PR #840, opened 2021-12-10
Replaced toolbar icon letters with square dots

cgorringe:
After review with the UI team, we decided to replace the letter inside the toolbar icons with dots. I initially tried a white round dot but found colored square dots looked nicer to me. I'm not sure if adding these dots to the buttons in the popup will be necessary so am not adding that for now.


infojunkie:
I reviewed the various testing conditions at TESTING_GUIDE. Some comments here:

- The count of saves hides half of the dot (which is already small)
- The color contrast between the dot / colored outline and the background is hard to distinguish

I suggest the following:
- Move the dot to the top-right corner of the icon and make it round, which is the expected location and shape 
- Increase the contrast of the blue and yellow colors used for the dot / outline
- Remove "Fact-checks" from TESTING_GUIDE

Alternatively, you could consider using a single, standard dot color and adding the dot to the relevant buttons in the extension popup, as was originally suggested.

cgorringe:
Here are what the toolbar icons & badges look like in various browsers, in Dark and Light modes on macOS Big Sur:

| Chrome | Firefox | Safari |
| --- | --- | --- |
| chrome-toolbar-icons | firefox-toolbar-icons | safari-toolbar-icons |

> Move the dot to the top-right corner of the icon and make it round, which is the expected location and shape

I understand what you are saying, but as you can see above, moving the dot to the top-right corner may look fine in Chrome, but would be completely covered in Firefox and Safari.

> Increase the contrast of the blue and yellow colors used for the dot / outline

I don't understand what you mean by increasing the contrast. It's already high between the outline color and the black background. I made the center dot slightly whiter than the outline as it looked more pleasing to me than the exact same color.

Safari will apply its own bluish tint to the icons (as shown above), but I have no control over that.

> Alternatively to the point above, you could consider using a single, standard dot color and adding the dot to the relevant buttons in the extension popup, as was originally suggested

I had initially tried a bright white center dot, but the colored one made the different colored icons easier to distinguish between them.

I can still add a square dot to the relevant buttons.

> Remove "Fact-checks" from TESTING_GUIDE

Done in #844

infojunkie:
> I understand what you are saying, but as you can see above, moving the dot to the top-right corner may look fine in Chrome, but would be completely covered in Firefox and Safari.

That's a bummer ! However, I still maintain that the color, shape and location of the dot that is proposed in this PR are not conveying the desired meaning. To convery a sense of notification, these dots are typically:

- using a bright color (which our blue isn't)
- placed overlaid at an icon corner (to convery that this is an add-on to the original icon, not part of it)
- of round shape (because that's the convention we see all over the web) 

My only suggestion to accommodate both the counts and the dots is to create versions for different browsers. 

I also suggest to run this PR by other UI-minded team members, e.g. Jim S. to get a second opinion.

cgorringe:
Fixed in #846

Issue #837, opened 2021-12-08
Remove Alexa Internet feature


cgorringe:
Alexa Internet will not longer be available as it is retiring May 1, 2022
https://support.alexa.com/hc/en-us/articles/4410503838999

PR #843, opened 2021-12-15
Replace Alexa Internet with My Archive

cgorringe:
Since the Alexa Internet service will retire in May (see #837), this PR removes that button, and replaces it with a new feature.

The Save Page Now website includes the option to “Save also in my web archive” which is also available on the iOS native app, but wasn't present in this web extension. So I added it as an option to the General Settings so that websites saved using the "Save Page Now" button can automatically be added to the user's web archive list. Then I added a new button on the main popup ("My Archive") which will open the user's page that displays this list.

|    |    |
|---|---|
| my-archive-ss1 | my-archive-ss2 |


infojunkie:
Works for me:

- Enable option "Save To My Web Archive"
- Visit page https://web.archive.org/web/20211216195614/https://www.discogs.com/release/21216217-Roland-Bourgeois-Miles-Black-David-Rehorick-Canadian-Horizons-A-Jazz-Collaboration (which had 0 save count so far)
- Click "Save Page Now" and wait until notification that saving is done
- Visit My Web Archive https://archive.org/details/@karim_ratib?tab=web-archive and see the URL

The only caveat is that My Web Archive shows a thumbnail of the site's front page, not the specific page I saved, as per screenshot. 

screenshot-archive org-2021 12 16-12_01_44

cgorringe:
Glad it worked. The thumbnail in My Web Archive is due to how the Wayback Machine saves screenshots and nothing can be done about it on the web extension end.

Issue #709, opened 2021-02-24
Blacklist for Auto-Save


s-crypt:
**Is your feature request related to a problem/ issue? Please describe.**

- There needs to be a blacklist to exclude auto saving sites. _**Private URLs and searches**_ can spam the WBM.

**Describe the solution you'd like**

- An implemented blacklist that has both defaults (like search engines. There is already a basic list [here](https://github.com/kissarat/never-lose/tree/master/chrome/list) and user added domains to not auto archive.

**Describe alternatives you've considered**

- Besides doing everything manually and/or turning on private mode, there is no other option.

**Additional context**

- It might also be a good idea to not archive the Alexa top 100-500 because they are already pretty much covered by manual archives, as well as pattern matching a url for a search query.

Edit: Some things that should be blacklisted:

- Anything with `search?`, `?s`, `s?`, `?searchterm`, `?search`, `?q`, `?query`, `search.cfm`, `search?q`, `result?`, `?st`, `search.aspx?`,  `*/search/*`, `*.amazon.com/s/`, `search#`, `?f`, `?t`  - used for search terms, can still be manually archived
- Delete anything after `?utm_source` - used for attribution. Remove anything after to get just the pure URL
- Delete anything after `#` - used for document locations

cgorringe:
This is a good point, particularly while using search engines and the `?utm_source` or that one that facebook uses.

s-crypt:
The [ClearURL extension](https://github.com/ClearURLs/Addon/) project actually has a list of regex matches to detect and make clean URLs.
https://docs.clearurls.xyz/latest/specs/rules/
https://rules2.clearurls.xyz/data.minify.json

cgorringe:
@s-crypt Thanks for the info!

cgorringe:
@s-crypt Really appreciate the suggestion and info on ClearURLs! I've created a new feature request at #722 - Thanks!

cgorringe:
Adding some notes on terminology for future reference:
- avoid: black / white.
- use: exclude, block, reject, minus.
- use: include, allow, permit, plus, safe.

"whitelist to exclude auto saving sites" actually sounds like the opposite. Maybe we can call this an "Exclude URL Patterns" list of regex's, or how about "Filtered URL Patterns".

s-crypt:
Sorry, I mixed up blacklist and whitelist when I originally wrote this

weirdalsuperfan:
I WANT THIS!!!

[...]

PR #845, opened 2021-12-22
Exclude URL List for Auto-Save

cgorringe:
Implements feature request #709 by adding a "Exclude URLs..." button to the Auto Save Page setting, which opens a window to allow the user to add their own list of URL patterns that will be skipped when the Auto Save Page setting is turned on. URL patterns support the * wildcard.

This is a really important new feature for Auto Save, as you can now keep it turned on without it trying to archive while reading GMail or using Google search, for example.

|   |   |
|---|---|
| exclude-urls-setting | exclude-urls-window |

Suggestions for additional URL patterns for the Defaults List are welcome!


infojunkie:
Thanks. Can you suggest a URL pattern that is guaranteed to NOT exist in the Wayback Machine? E.g. a site that generates random URLs. This way we can test that excluded patterns do work.

PR #846, opened 2021-12-30
Replace Toolbar Icons with Notification Dots

cgorringe:
2nd attempt at updating the toolbar icons. This includes code that will switch the set of icons depending on the browser so that the notification dot isn't covered over by the Wayback Count badge. Includes adding round dots to buttons and settings.

See #840.

### Toolbar Icons
- Safari will tint the red dot blue, which we can't do anything about.

| browser | toolbar icons |
| --- | --- |
| Chrome | chrome-toolbar-icons-take2 |
| Firefox | firefox-toolbar-icons-take2 |
| Safari | safari-toolbar-icons-take2 |

### Chrome Screenshots
- "Cited Books/Papers" is renamed to "Books/Papers" to fit the dot in.
- 404 Error Message color changed from orange to red.

|  |  |  |  |
| --- | --- | --- | --- |
| books-papers | 404-not-found-dark | 404-not-found-light | settings |


infojunkie:
Thanks, this is much clearer signaling. I've tested the various conditions on Chrome and they look good to me :tada:

webextension/scripts/utils.js
```diff
@@ -116,7 +116,7 @@ if (typeof isInTestEnv === 'undefined') {
  */
 function isBadgeOnTop() {
   // TODO: check other browsers
-  const badgeOnTop = { firefox: true, safari: true, chrome: false }
+  const badgeOnTop = { firefox: true, safari: true, chrome: false, edge: false }
```
infojunkie, line 119: I suggest to ensure that a default ia returned if the browser key is not mentioned.

cgorringe: It's intentional to return `undefined` with unknown browsers. Works like false when in an if expression. Means we don't know what the correct value is yet.

infojunkie: As long as the extension doesn't crash and shows some icon, even if not the best one, that's fine.

cgorringe: Of course! JS ain't Java or C! And I tested this, it works.

PR #854, opened 2022-01-28
Revive 404 popups

cgorringe:
This adds back the html-embedded 404 popups, but keeps the new style 404 errors, while only turning on the old-style popups when an option is turned on in Settings.


infojunkie:
Thanks, works for me with the popup setting turned ON and OFF. Also help text looks good to me.

PR #865, opened 2022-02-22
Remove threshold value comparision in tvnews.js

anishsarangi:
For now we are removing the threshold value condition in tvnews.js. We will add this filtering in the `/tvnews` backend.


infojunkie:
Looks good to me.

infojunkie (approved):
Looks good to me

PR #881, opened 2022-03-01
Fix popup not showing in firefox when tabID is -1

### LINKED ISSUE #878: Embed 404 popup doesn't always display in Firefox
**Describe the bug**

When on an error page such as a 404, the red dot and popup error will display, but the embedded html popup doesn't always display, but will display after website is reloaded.

**To Reproduce**

- Use Firefox.
- Turn on "Replace 404s..." and "Errors Pop-Up in Page" in Settings.
- Go here: https://www.whitehouse.gov/energy/climate-change

**Expected behavior**

- Html popup should display first time, every time.

**Environment**

 - Webextension version: 3.0
 - Browser + version: Firefox
 - OS + version: macOS

anishsarangi:
Fixes #878


infojunkie:
Thanks @anishsarangi - can you please explain how the fix solves the issue? I.e. how does `/scripts/archive.js` help?

anishsarangi:
@infojunkie So when you paste the url in firefox browser, `chrome.webRequest.onCompleted` event is triggered. But as a bug in firefox, we are getting tabId as -1. So in that case we are using `chrome.tabs.query` to get the tabID and after that we execute the script archive.js in that tabID.  After that in the `update` function, we send a `SHOW_BANNER` message.Then 
 the popup appears in the tab.

cgorringe:
@anishsarangi It's not quite correct yet, but it is the right idea! Try doing this over again moving `chrome.tabs.query()` outside of `checkWM()` and make sure that `executeScript()` never gets called when bannerFlag is false.

cgorringe (approved):
Works fine now. Please fix spacing & indentation then you can squash/merge to master.

cgorringe:
Sorry, actually there's still an issue with inserting the script when bannerFlag is false which I missed before. I'll just fix this...

Issue #873, opened 2022-03-01
Some observations on TV News Clips


cgorringe:
**Describe the bug**

1. While trying out TV News Clips, it seems like some of the screenshots don't really match the videos they link to, although sometimes they do, but other times it seems like they match one of the other videos.

2. The link URL that the browser shows is the same for every video, even though it properly goes to different video links after clicked, it would be nice to see the actual link when hovering over the screenshots.

3. When clicking on a video, the TV News Clip window automatically scrolls to the top, which is odd.

**To Reproduce**

Turn on TV News Clips then go to this URL and tap the blue button once it shows up:
https://edition.cnn.com/europe/live-news/ukraine-russia-news-02-23-22/h_6d205f365f501bc935268528b826a824

**Screenshots**

![Screen Shot 2022-03-01 at 12 42 26 AM](https://user-images.githubusercontent.com/604259/156136103-137bf1ca-050e-48e6-beb0-6fd82701774f.png)

**Environment (please complete the following information):**
 - Webextension version: 3.0.0.3
 - Browser + version: Chrome 98
 - OS + version: macOS 11.6.3

cgorringe:
Anish noticed that the blue dot and the button is appearing after the url is completely loaded. (maybe we can speed this up?)

cgorringe:
All but (1) fixed.

PR #882, opened 2022-03-01
Show tvnews button when the url loading is in progress and show href links on hover

anishsarangi:
Fixed 2nd and 3rd point from #873 .
Also fixed the speed issue while showing tvnews button in the ui


cgorringe (approved):
Tested on Chrome and reviewed the code. Looks good. Covered 2, 3, and 4. Blue dot displays faster - great.

PR #884, opened 2022-03-04
Fix Site Map button while in Word Cloud window

### LINKED ISSUE #874: Checking buttons work in some cases
**Describe the bug**

From Mark:
- If I select “Word Cloud”, then select “Site Map” my request is not honored (and the button does not change color).  If I select “My Archive” and THEN “Site Map” my request IS honored.

**Environment**
 - Webextension version: 3.0
 - Browser + version: Safari
 - OS + version: macOS

cgorringe:
fixes #874


infojunkie:
Thanks, I verified it's broken on the release version and fixed here.

infojunkie (approved):

Issue #880, opened 2022-03-01
Show the development icon differently than the release one


infojunkie:
When working on this extension, we sometimes need to keep both the development version AND the latest release present in the toolbar for comparison and bug tracking. It would be great to be able to quickly differentiate between them, e.g. by coloring the dev version icon differently.

infojunkie:
Thanks, that would work

infojunkie:
When the red dot is shown, the icon is white instead of cyan.

cgorringe:
Yes, it's only the main icon, not any that have dots.

infojunkie:
> Yes, it's only the main icon, not any that have dots.

Thanks for confirming - but I think this adds to the confusion, not reduces it.

cgorringe:
It's not like I set out to do it. I have to add more icons to fix it. But I'll see, it may be a quick enough fix.

infojunkie:
Thanks, looking good now!

PR #885, opened 2022-03-04
Cyan-colored toolbar icon for beta dev builds

cgorringe:
implements #880

Changes main archive icon to cyan for versions that have 4 numbers (x.x.x.x) which are development builds.


infojunkie (approved):
Thank you, looks fine.

Issue #867, opened 2022-02-24
Feature: Context Notices


infojunkie:
Add a notification dot when a Context Notice is found for the current page. What's a Context Notice? See this [blog post](https://blog.archive.org/2020/10/30/fact-checks-and-context-for-wayback-machine-pages/) for background, and [this example](https://web.archive.org/web/20220127123423/https://twitter.com/JesseKellyDC/status/1463280652418228229).

Description of the feature flow:
- User visits a page
- Extension queries backend for existing notices matching the URL 
- Backend answers with list of notices and optional timestamps where the notices appear on Wayback Machine
- In case of found notices, display a notification dot (TBD which color? if different than blue, find a way for multiple dots to co-exist)
- User opens the extension popup
- Extension shows an additional button (e.g. "Contextual Notices") that links to the corresponding capture on Wayback Machine - either the timestamp capture, or the latest one if no timestamp available. In case of multiple notices + multiple timestamps, pick the most recent timestamp.
- Add a checkbox for the feature in "Context" tab labeled "Contextual Notices" with help text "Check for contextual information from fact checking organizations and origin websites" and showing yellow dot. This setting is OFF by default.

infojunkie:
In case of multiple notices + multiple timestamps, pick the most recent timestamp.

infojunkie:
Added option setting for the feature in the description.

PR #887, opened 2022-03-05
Context Notices feature

cgorringe:
implemented #867


infojunkie:
Thanks, I checked on URLs that have notices and I see the yellow dot, the notices button, and it redirects me to the relevant timestamp :tada: I'll be testing more over the week.

infojunkie:
Thanks, the latest commit fixed the condition of empty where/timestamps. Keeping this PR open until the Nomad backend is ready.

Issue #888, opened 2022-03-05
Why is it required to be logged in to archive a page?


mitar:
Previous version of the extension on Firefox did not require that. Doing archive through website itself also does not require that. Why is this required for the new version of the extension? I understand that it is required to access additional features, but for basic archiving it should work without being logged in also through the extension.

cgorringe:
We are working on getting this to work.

cgorringe:
Thank you, then next version 3.1 will let you archive while logged out.

Issue #872, opened 2022-02-26
Log in request


krisszt:
**Is your feature request related to a problem/ issue? Please describe.**
When the prompt suddenly came up to log in today through a notification I was very confused on why, and even where I would log in.

**Describe the solution you'd like**
Is it possible to make it so that a click on the notification takes the user to the archive's [log in page](https://archive.org/account/login)?

**Additional context**
I already had an account there, the sudden need to log in there to use the extension just didn't click for a bit. I assume it will be an issue to other users too.

(Even if I don't understand the why of it.)

cgorringe:
If you click on the archive toolbar button for the popup, then move the mouse pointer over the "Save Page Now" button while logged out, it will flip over to show "Log In to Save Page", which you can then log in right inside the extension. But we can also see about adding a link to the archive's log in page to the notification. Thanks for the suggestion!

Issue #864, opened 2022-02-19
Disable requirement to log to archive.org account prior to enabling "Save Page Now"


cstanhope:
It appears that with the 3.0 version of the Wayback Machine extension that you must now log in to your archive.org account before the "Save Page Now" button will be enabled. This wasn't the case in prior versions. I don't know the full reasoning behind this change other than I note there are at least a couple of new options (off by default) that would probably require a person to be logged in. One of these is "Save to my Web Archive" and the other I can see is "Email Outlinks". There might be others I'm missing.

I don't ever want to save anything I've archived to my account, and I'm not interested in any of the other personalization features. I'm also one of those people that frequently clears browser data and ends up logged out of everything, so I will be logging in repeatedly to use the "Save Page Now" button, which eliminates the efficiency it would otherwise provide. 

Additionally, I use this browser extension in contexts where I do not wish to log in to my account (e.g. at work). In those contexts, the "Save Page Now" feature is now unusable.

**Describe the solution you'd like**
It would be nice if the "Save Page Now" button was enabled even when logged out if none of the personalization features were enabled. 

Even if this is possible, I admit that this appears to be a UX challenge. Right now the UX is the same for everyone, but if my change could be implemented, the experience for people who have enabled personalization functions will be different from those that haven't. I don't necessarily have a good suggestion to overcome that other than to provide additional context as to why the user must log in. 

**Describe alternatives you've considered**
If this can't be made to work, my work around would be to stop using the extension for saving pages and go back to using the web interface.

cgorringe:
Thanks and we appreciate your feedback! I can see why you may not want to have to log in to use Save Page Now. Unfortunately it's not something that I can change back (at least right now) but I _might_ be able to keep it logged in while clearing the browser's cookies so that you wouldn't have to repeatedly log in. I like to clear my browser cookies too. Seems like a common situation. Thanks for pointing it out! Will see what I can do.

cstanhope:
Great! Thanks! And this may not be the place for this comment, but since I'm here, I'd like to say I really like and appreciate the other recent changes to the extension.

MurphysLab:
> Unfortunately it's not something that I can change back (at least right now)

Would it be possible to make the previous version of the browser extension available separately? 

Some do appreciate the privacy aspect of not having our archiving activity tracked. I hope that this gets changed so that login isn't required.

Also, can anyone here explain why this design **choice** was made?

rsaarelm:
[...]

PR #891, opened 2022-03-10
Save Without Login

cgorringe:
A solution for #864 #872 #888

- The Save Page Now button and right-click Save menu can now work while logged out. 
- It opens a new website page in order to do so, and can't do the usual UI animations. 
- <s>Dims and disables Settings which relate only to the logged-in version of Save Page Now. </s>
- Checks for login state during Auto Save to fix a prior bug.
- Clicking on options or settings will switch to Login view if not supported when logged out.

alt-login-view


cgorringe:
v3.0.0.10 merges in Context Notices, and a 404 popup bug fix.

PR #898, opened 2022-03-15
Save Without Login using a form redirect

cgorringe:
improving #891
Opens a Wayback Machine webpage to Save Page Now anonymously when SPN button pressed.

Since there's no direct way to POST to a URL inside web extension code, I had to open a temporary internal page to automatically submit a form while passing in the URL to save.


infojunkie:
It works great! Because the intermediate redirection page may show for a few seconds, I suggest to rename the text there "Redirecting to Save Page Now..."

infojunkie (approved):

infojunkie (approved):

PR #899, opened 2022-03-16
Fix for settings reset and welcome redisplay on upgrade

### LINKED ISSUE #890: All settings are being reset randomly
Extension settings are being reset to default randomly. My guess is this is the cause:

https://github.com/internetarchive/wayback-machine-webextension/blob/c7dddac4cf72758e75b30d80e980c91567159a03/webextension/scripts/background.js#L387-L388

According to https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onInstalled, the event is
> Fired when the extension is first installed, when the extension is updated to a new version, and when the browser is updated to a new version.

So settings will be reset after extension or browser update. I can't tell if this is the case because I only use the extension occasionally.

The fix would be to check the details.reason object passed into the callback. https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/OnInstalledReason

 

 - Webextension version: 3.0
 - Browser + version: Firefox 98.0
 - OS + version: Windows 10

cgorringe:
fixes #890

### DIFF
--- a/webextension/scripts/background.js
+++ b/webextension/scripts/background.js
@@ -5,7 +5,7 @@
 
 // from 'utils.js'
 /*   global isNotExcludedUrl, getCleanUrl, isArchiveUrl, isValidUrl, notify, openByWindowSetting, sleep, wmAvailabilityCheck, hostURL, isFirefox */
-/*   global initDefaultOptions, afterAcceptOptions, badgeCountText, getWaybackCount, newshosts, dateToTimestamp, fixedEncodeURIComponent, checkLastError */
+/*   global initDefaultOptions, badgeCountText, getWaybackCount, newshosts, dateToTimestamp, fixedEncodeURIComponent, checkLastError */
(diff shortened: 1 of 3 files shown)


infojunkie:
How do yo usuggest to test this fix?

cgorringe:
Good point. You might not be able to test all of it, but one thing you can test is:
1. Remove / uninstall the extension, then reinstall. Should result in welcome screen when tapping the toolbar icon.
2. Turning the extension switch on / off, or reload, should no longer cause the welcome screen to reappear.
3. The welcome screen should also not reappear when updating the browser, but likely can't test this until deployed.

Or if it looks good I could just merge it.

I tested the above on Chrome & Firefox. Will test again on Safari. Hadn't tested on Edge.

infojunkie:
Thanks, I tested the 3 conditions mentioned and they worked as advertised.

infojunkie (approved):

PR #900, opened 2022-03-16
Welcome page buttons always visible

cgorringe:
We noticed that some users missed the Accept button in the Welcome screen, which would show repeatedly since they never clicked on the Accept button before closing the window. This PR places the buttons fixed along the bottom edge so that they're always visible.


infojunkie (approved):
Beautiful

PR #903, opened 2022-03-16
Move the .click event out of setupSaveAction which was triggering on every input box change event

### LINKED ISSUE #901: Save Page Now is opening 6 tabs instead of 1 when logged out
**Describe the bug**
If a user is not logged in and when it try to save a url through the input box in the extension ui, it is opening 6 tabs simultaneously.
**To Reproduce**
1. Go to any URL
2. Click on the extension and make sure that you are logged out
3. Type any url in the extension and click "Save Page Now"
4. You will see 6 tabs opening simultaneously

**Expected behavior**
It should open 1 tab

anishsarangi:
Fixed #901


cgorringe (changes requested):
I don't think this is correct. setupSaveAction() probably shouldn't be called repeatedly but only once? But to fix just that one line this will likely work: (need to test)
`$('#spn-btn').off('click').on('click', doSaveNow)`

anishsarangi:
`setupSaveAction` is called repeatedly to get the LastSaveTime of the url which was in the input box.
But `$('#spn-btn').on('click', doSaveNow)` will setup the event which is called on clicking on `#spn-btn`

infojunkie:
Fix works and looks good to me.

cgorringe:
This code introduces another issue: While logged in, click Save Page Now while on an excluded URL like `chrome://extensions/` causes SPN button to animate. My suggestion should fix the immediate issue above.

Another issue with the search box is that it will call the sparkline API multiple times while typing a URL, once for every key pressed while typing. I think it may be because `useSearchBox()` is being called on every key press event, which is running a bit of code which shouldn't be repeatedly. @anishsarangi do you want to fix this?

cgorringe:
Since PR #915 replaces the branch this is based on, this no longer applies.
Search box issue moved to #909.

cgorringe:
Try going to a Wayback URL, then tap SPN button.

cgorringe:
fixed in #919

PR #904, opened 2022-03-16
Reset my archive settings after reloading the extension

anishsarangi:
Reset my archive settings after reloading the extension to false by default.


cgorringe (approved):
I must have missed this. Thanks for catching it!

PR #915, opened 2022-03-23
SPN Without Login with Full UI

### LINKED ISSUE #905: Save Without Login fails in Safari
**Describe the bug**

When trying Save Page Now in Safari, while logged-out, the spn-redirect.html and .js code does redirect to the correct website, but the URL doesn't get passed in to fill in the form text box. Works fine in Chrome & Firefox.

**To Reproduce**

1. Go to any URL while using Safari.
2. Make sure you're logged out.
3. Click on red Save Page Now.
4. See redirection to https://web.archive.org/save website missing the URL in text box.

 - Webextension version: 3.0.0.13
 - Browser + version: Safari 15.3
 - OS + version: macOS 11.6.3

**Additional context**
I've been trying to fix this bug without any luck so far.

cgorringe:
Fixes #905 and replaces PRs #898 and #891

- Solves Save Page Now while logged-out, including UI such as toolbar icon and animated button.
- Enables "Show Resources During Save" and "Auto Save" settings now work while logged out.
- Removes the spn-redirect.html and .js files, as the alternate website in a window solution is no longer needed.


cgorringe:
Auto Save broken in Safari between master (v3.0.0.13) and this PR. Will fix...

infojunkie (approved):
Since the code paths for logged in and logged out are now merged, except for a few differences, I tested both cases and they worked fine, with and without the resources dialog option. Great stuff!

PR #918, opened 2022-03-26
optimizes, shrink rasters

mathieu-aubin:
please feel free to close

suggestion:

youtube.com and youtu.be in ignore url list


cgorringe:
Thanks for the youtube in ignore URL list suggestion.

I'm going to close this without merging because we don't really want to optimize every graphics file, only the ones in the actual extension under the `webextension` directory. And it would be great if we had a script or instructions to repeat the process in the future.

mathieu-aubin:
Another suggestion as far as ignore list... i forgot youtube-nocookie.com
Try and find, as accurately as possible, the major FINANCIAL institutions webpages... Tangerine, RBC, Desjardins (quebec, Canada) ... etc
netflix, nebula, full30, curiositystream .... there will always be more i get it but the most obvious ones would go a long way
online gambling sites... casinos...

Also... There is a way to automate this raster optimization somehow. If you wanted, i could give it a go and present the easiest best solution i find.

You are FREE to use the compressed graphics from this pull request

PR #923, opened 2022-03-29
Stop making amazonbooks api call in every url containing amazon

anishsarangi:


cgorringe (approved):
This looks good. We should add some caching to these API calls in a future PR.

infojunkie (approved):

PR #926, opened 2022-03-31
Replace checkboxes with login icons when logged out

cgorringe:
This solves a UI issue that was bugging me where you wouldn't know what checkboxes sent you to the login screen when logged out. This also adds a login button icon to the lower-right of the main popup screen.

spn-mini-login-icons


infojunkie (approved):
Extremely pleasing interaction! Thanks!!

PR #938, opened 2022-04-21
Check the username when the login get succeed

### LINKED ISSUE #937: When logged out from the site, and logged back in from the site from other account, My Archive redirected me to the previous account
**Describe the bug**
When the user logs out from it's account in [archive.org](http://archive.org/) website, logged back in from the site from other account and click on the "My Archive" button, the user name is redirecting to the previous account

**To Reproduce**
1. Log into the [archive.org](http://archive.org/) with one account and click on "My Archive" button.
2. Then logout from archive in the [archive.org](http://archive.org/) website and login with other account.
3. Click on "My Archive" button.
4. This will open the previous account's "my web archive" page.

**Expected behavior**
The user's id should change on changing the account in the website.

anishsarangi:
Fixed #937

### DIFF
--- a/webextension/scripts/popup.js
+++ b/webextension/scripts/popup.js
@@ -208,20 +208,6 @@ function loginSuccess() {
   $('#login-tab-btn').hide()
   $('#logout-tab-btn').css('display', 'inline-block')
 
-  getUserInfo().then(info => {
-    if (info && ('screenname' in info) && ('itemname' in info)) {
-      var previous_user_itemname = ''
-      chrome.storage.local.get(['itemname'], (settings) => {
-        if (settings && settings.itemname) {
-          previous_user_itemname = settings.itemname
-        }
-      }) 
-      if (previous_user_itemname != info.itemname){
-        chrome.storage.local.set({ screenname: info.screenname, itemname: info.itemname })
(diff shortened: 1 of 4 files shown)


webextension/scripts/popup.js
```diff
@@ -208,6 +208,20 @@ function loginSuccess() {
...
 
+  getUserInfo().then(info => {
+    if (info && ('screenname' in info) && ('itemname' in info)) {
+      var previous_user_itemname = ''
+      chrome.storage.local.get(['itemname'], (settings) => {
+        if (settings && settings.itemname) {
+          previous_user_itemname = settings.itemname
+        }
+      }) 
+      if (previous_user_itemname != info.itemname){
```
cgorringe, line 219: This if statement is not guaranteed to run after the previous assignment to `previous_user_itemname` since that assignment is in an async block.

cgorringe (changes requested):
Since loginSuccess() is called every time the user taps on the toolbar icon while logged in, the call to `getUserInfo()` would make an API call every time. This seems excessive. Also, there is no check for Private Mode, so would violate privacy.

cgorringe (approved):

Issue #897, opened 2022-03-15
Checking for notices despite unreachable domains and other network errors


infojunkie:
A user reported that for https://ria.ru/20220226/rossiya-1775162336.html, they got an error that the site could not be reached `DNS_PROBE_FINISHED_NXDOMAIN`. As a result, the extension reported "Server Not Found" and did not attempt to check for notices. On the other hand, the 404 handler worked fine.

![domain](https://user-images.githubusercontent.com/92977/158285075-c3720039-f508-40ac-9fc4-8f9f1e93762b.png)

The extension should still check for notices in this case.

cgorringe:
`embed_popup_setting` is the option in Settings to turn on or off the html-embedded 404 banner. When it is set `false`, the banner should not display, only the toolbar icon and popup should indicate when a 4xx/5xx error occurs.

VishnuRaghavan:
@cgorringe  Oh I see, could you explain what does "should check for notices" mean as @infojunkie mentioned in above comment. From what i understood WM should check for saved versions of the url if it couldn't resolve it as in this case of 404 ?. Is that correct ?

infojunkie:
In the case I reported above, the error condition was not 404, but unreachable domain.

cgorringe:
The "Server Not Found" error does look correct. I think the issue was that the "Contextual Notices" yellow button should also display (if turned on) and wasn't being displayed on Server Not Found errors. For me, when I check that URL, I get a 404 Not Found and the "Contextual Notices" does correctly show, so I haven't reproduced the issue on that particular URL.

PR #942, opened 2022-05-04
Fix for Issue #897

VishnuRaghavan:
Once wayback machine returns status code 404 a popup should be displayed to view older saved versions. This particular fix is for the same.


cgorringe:
`embed_popup_setting` is the option in Settings to turn on or off the html-embedded 404 banner. When it is set `false`, the banner should not display, only the toolbar icon and popup should indicate when a 4xx/5xx error occurs.

Issue #912, opened 2022-03-21
Disable popup buttons when 'URL not supported'


cgorringe:
We may want to dim or disable the buttons in the popup when "URL not supported" text is displayed, since currently (v.3.0.0.13) when user clicks on a button, none of the buttons do anything, including the social share buttons. Except the SPN button currently flips to "URL not supported" when hovered over, which is fine.

VishnuRaghavan:
Will look into this.

VishnuRaghavan:
I have worked on this enhancement, please check below screenshot. Let me know if any suggestions.

**When URL is not supported** 

![Screenshot 2022-05-06 at 11 18 52 PM](https://user-images.githubusercontent.com/23146562/167203769-0df28bae-55ee-4a1a-8987-c0c33f0f7cc2.png)



**When URL is supported**

![Screenshot 2022-05-06 at 11 20 53 PM](https://user-images.githubusercontent.com/23146562/167203932-8464fa5a-1260-40f9-9a70-c381e0baf16f.png)

cgorringe:
I'd say "Outlinks" & "Screenshot" shouldn't be dimmed.
Also "My Archive" since it doesn't make use of the current URL.
Everything else makes use of the URL so look correct.

cgorringe:
Also test with the Search box, since the current URL would change and thus have to update the buttons dim state.

cgorringe:
Also Thanks for working on this @VishnuRaghavan !

VishnuRaghavan:
@cgorringe  Will work on this.

VishnuRaghavan:
@cgorringe  Hey i have made above mentioned changes, check this PR #943

Issue #913, opened 2022-03-21
Make popup buttons accessible using tab key


cgorringe:
Currently (v3.0.0.13) some buttons on the popup can't be reached using the tab key, including 'Save Page Now', which might require removing the hover flip. (but hope we can keep the flip.) It would be great if all the buttons and checkboxes can be reached using tab.

VishnuRaghavan:
I have worked on tab navigation. i will create a PR please do check. #943

cgorringe:
Thanks @VishnuRaghavan I will try it out, but the code so far looks great!

PR #943, opened 2022-05-06
Enhancement/popup btn tab fix

VishnuRaghavan:
Added tab navigation for popup buttons.
Replaced stock blue outlines of buttons with its own background selected colour while on focus.


cgorringe:
Thanks for these great improvements!

Some items to improve further:
- Let's see if we can keep the UI the same for the Save Page Now button with options underneath without adding space in between.
- In Dark mode, simple tweak needed for SPN button background color which shows up as white corners.
- Pressing Space or Enter on Outlinks, Screenshot, Social share buttons, and bottom toolbar buttons should activate their function.
- First item focused could be the Search field instead of the top logo.

VishnuRaghavan:
Is it okay to make above mentioned changes on to same branch ?

VishnuRaghavan:
> * In Dark mode, simple tweak needed for SPN button background color which shows up as white corners.

Which browser have you used to test, im using mac os i can't see white corners around SPN.  Im attaching a screen shot please check.

Screenshot 2022-05-07 at 1 18 01 AM

cgorringe:
It's fine to use the same branch!

I'm getting the white corners only when "URL not supported" and when the button flips. I'm using Chrome on macOS.

white-corners

VishnuRaghavan:
@cgorringe okay let me check

VishnuRaghavan:
> * Pressing Space or Enter on Outlinks, Screenshot, Social share buttons, and bottom toolbar buttons should activate their function.

For this particular fix, we might need to listen to keyup and keydown events along with key-codes. which will definitely  fix it for all browsers. I see chrome had an issue with keys not detecting during tab navigation. also sometimes its doesnt work for span elements and divs. So im gonna listen to event and take action from there.

VishnuRaghavan:
I might need to change all popup buttons to actual button tag instead of other elements like span and div. button tag will properly work with tab navigation. This is one approach, if this cant be achieved then i would go with listening to key codes on key down events as i mentioned above.
But cleanest approach it to use button tag, let me try that first.

VishnuRaghavan:
@cgorringe  Changes mentioned above has been added, please do check.

cgorringe:
@VishnuRaghavan I love how it's now working using the keyboard. Nice!

So far I found these issues:
- Activating "Save Page Now" button ("Archiving...") no longer animates, which was a pretty cool effect we'd want to keep!
- While logged in, try clicking on "Outlinks" or "Screenshot" labels with the mouse - it will no longer check or uncheck the box.
- Also these labels get stuck highlighted after a mouse click.

Thanks for your work so far. I know it takes a while to get everything working just right.

VishnuRaghavan:
Will check that.

VishnuRaghavan:
> @VishnuRaghavan I love how it's now working using the keyboard. Nice!
> 
> So far I found these issues:
> 
> * Activating "Save Page Now" button ("Archiving...") no longer animates, which was a pretty cool effect we'd want to keep!
> * While logged in, try clicking on "Outlinks" or "Screenshot" labels with the mouse - it will no longer check or uncheck the box.
> * Also these labels get stuck highlighted after a mouse click.
> 
> Thanks for your work so far. I know it takes a while to get everything working just right.


@cgorringe  I have fixed above mentioned issues please do check.

cgorringe:
Sorry for the delay. I tested and made some changes, reverting the last commit as the additional code made it too complex and wasn't really needed. 

I didn't succeed in getting the background to highlight while using the keyboard just using CSS, but that really wasn't needed and it's more important for the mouse UI to actually look correct while using the keyboard just needs to be functional.

cgorringe:
The jQuery `.click()` actually isn't deprecated according to the official docs, although I did find mention of it on the net. Not sure what happened. Maybe they deprecated it at one point then changed their minds? So I'll revert those lines back next...
https://api.jquery.com/click/

cgorringe:
This PR fixes #912 and #913
Except Safari extensions don't seem to work using the keyboard, but not much we can do about that.

Issue #896, opened 2022-03-14
Opening the archived copy of a URL without creating a new tab


el4ctr0n:
Dear Wayback Machine Chrome extension developers, can you please make it possible to configure in the settings that when I press "Oldest", the oldest archive of the URL would open in the same tab? 

This feature would be enormously helpful for the situations when you need archived copies of multiple currently open URLs. 

Thank you for considering my request. 
I look forward to your reply.

cgorringe:
Thanks for the suggestion!

We can consider changing the setting for "Show Features In"
( ) Tab ( ) Window ( ) Replace [or something like that...]

el4ctr0n:
> Thanks for the suggestion!
> 
> We can consider changing the setting for "Show Features In" ( ) Tab ( ) Window ( ) Replace [or something like that...]

Can you please add this feature? Thank you for your time.

cgorringe:
Added in v3.1.0.3

PR #950, opened 2022-06-28
Show features in Same Tab option

cgorringe:
Implements #896

Adds 'Same Tab' option to Settings:
show-features-in

There's a reported bug in Chrome where the browser's Back button won't let the user go back to the prior page after opening a URL when Same Tab option is active. Works fine in Firefox.


cgorringe:
Tested in:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari

Issue #749, opened 2021-05-18
Auto-Save New Bookmarks


cgorringe:
This is an enhancement to the existing Auto Save feature, where an option is added to Auto Save whenever a user saves a bookmark. Perhaps add a checkbox in Settings.

Idea from @vbanos - thanks!

cgorringe:
See also #673 Auto-Update Previously Saved Pages.

anishsarangi:
Some resource: 
https://developer.chrome.com/docs/extensions/reference/bookmarks/#event-onCreated

nahoj:
I would also appreciate such a feature.

I don't want to use Auto Save because I don't want an extension to track every page I read, but whenever I bookmark a page it's typically somewhat important to me so I'd appreciate it if it was saved.

Upon failure the user could be notified and offered to ignore future bookmarks on that domain name (useful for intranets and websites that require authentication).

cgorringe:
@nahoj Thanks for your input! I think it would be a good addition.

PR #962, opened 2022-08-10
Auto Save New Bookmarks

cgorringe:
Implements #749 

- Auto Saves when user bookmarks a web page, without notification.
- Ignores excluded URLs including user-defined Exclude URLs.
- Uses optional 'bookmarks' permission to avoid warnings at install time, which means the browser will ask permission when user tries to check the setting for the first time.
- Can't save multiple tabs at once. Will only save the current tab.
- Doesn't attempt to save when importing a list of bookmarks.
- Hides the setting in Safari, since bookmark access not supported.
- Tested on Chrome, Firefox, Edge, and Safari.

auto-save-bookmarks


nahoj:
Thank you for developing this, it's cool!

Issue #911, opened 2022-03-21
Keep user logged in after clearing browser cookies


cgorringe:
Some folks like to clear their browser's cookies frequently. Unfortunately this will also log them out. (v3.0.0.13) We'd prefer, if it were possible, to keep the user logged in after clearing their cookies so that they wouldn't have to repeatedly log in.

See #864

PR #963, opened 2022-08-12
Stay logged in after user clears cookies

cgorringe:
implements #911

This fixes a common complaint when the user cleared their cookies and found themselves logged out. This PR will keep them logged in until they explicitly log out using the log out button in the extension. This works by storing a copy of the auth cookies in storage and restoring them after cookies were cleared and the user clicks the toolbar button.


cgorringe:
Code modified so that logging out from archive.org website will also logout from the extension properly, without accidentally restoring cookies that would cause the user to be logged back in when reloading the website, since they expect to be logged out.

anishsarangi:
Tested on Chrome, Firefox and Edge. It's working great. We can merge this.

cgorringe:
Tested and works on Safari, Firefox, and Edge.

PR #965, opened 2022-08-16
Prevent Auto Save from re-saving recently saved pages

### LINKED ISSUE #964: Auto save page constantly re-save page despite I choose "if not archived with past year"
**Describe the bug**

I set Auto save page on and condition is "if not archived: within past year". Yet, it constantly save the same page (at many as 5 copies just today).

**To Reproduce**
1. Enable Auto Save Page.
2. For "if not archived:", choose "within past year".
3. Open some pages.

**Expected behavior**
It should only archive it at the first sight (assuming not archived by others).

**Screenshots**

As you can see here:
image

I'm not going to post the exact URLs for now since it would doxx myself, but it happens on lots of websites. One example is Hacker News' user page: https://news.ycombinator.com/threads?id=XXXX (I'm sure it's saved by me since it's my account and there is zero saves before I enabled this feature).

 - Webextension version: 3.1
 - Browser + version: Chrome Version 104.0.5112.81 (Official Build) (64-bit)
 - OS + version: [Windows 10]

cgorringe:
Fixes #964

- Replaces Availability API call with cached WM count (sparkline) API call in autoSave().
- Tested on Chrome. Needs testing still on Firefox, Edge, Safari.

### DIFF
--- a/safari/Wayback Machine.xcodeproj/project.pbxproj
+++ b/safari/Wayback Machine.xcodeproj/project.pbxproj
@@ -461,7 +461,7 @@
 					"@executable_path/../../../../Frameworks",
 				);
 				MACOSX_DEPLOYMENT_TARGET = 10.14;
-				MARKETING_VERSION = 3.1.0.7;
+				MARKETING_VERSION = 3.1.0.8;
 				PRODUCT_BUNDLE_IDENTIFIER = archive.org.waybackmachine.mac.extension;
 				PRODUCT_NAME = "$(TARGET_NAME)";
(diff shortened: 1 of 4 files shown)


cgorringe:
Testing Safari: Auto Save failed to run when the Wayback Count (sparkline) API call timed-out, but later worked when the API responded quickly. If the Availability API has better response times, then this might fail more often than it used to.

cgorringe:
While testing Chrome: Clearing "Wayback Machine Count" in Settings will clear the cache, causing Auto Save to activate again when revisiting a page.

PR #976, opened 2022-12-13
Remove mastodon.archive.org from excluded urls

anishsarangi:
Make changes in `isArchiveUrl` function to include mastodon.archive.org as a supported url in the extension.

### DIFF
--- a/test/utils.spec.js
+++ b/test/utils.spec.js
@@ -24,9 +24,9 @@ describe('twitter', () => {
 
 describe('isArchiveUrl', () => {
   var test_cases = [
-    { 'url': 'http://archive.org', 'result': true },
-    { 'url': 'https://archive.org', 'result': true },
-    { 'url': 'http://archive.org/some/path/?key=value', 'result': true },
+    { 'url': 'http://archive.org', 'result': false },
+    { 'url': 'https://archive.org', 'result': false },
+    { 'url': 'http://archive.org/some/path/?key=value', 'result': false },
     { 'url': 'https://web.archive.org', 'result': true },
     { 'url': 'https://web.archive.org/some/path', 'result': true },
     { 'url': 'http://example.com', 'result': false },
--- a/webextension/scripts/utils.js
+++ b/webextension/scripts/utils.js
@@ -15,8 +15,6 @@ const defaultAutoExcludeList = [
 
 // list of excluded URLs
 const excluded_urls = [
-  'localhost',
-  '0.0.0.0',
   '127.0.0.1',
   'chrome:',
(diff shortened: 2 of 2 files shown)


vbanos:
@anishsarangi I see that we block `archive.org` and we allow `mastodon.archive.org`.

Allowing/Blocking a URL is happening in SPN, it shouldn't happen in the browser extension.
You should allow all URLs and the SPN will decide to capture it or not.
(allow/block has so many rules, e.g. we also also blog.archive.org and help.archive.org).

So, please drop `isArchiveUrl` altogether. Thanks.

anishsarangi:
@vbanos We are showing based "URL not supported" on the UI based on `isArchiveUrl`, `isNotExcludedUrl` and `isValidUrl`. How can we show it without storing urls? Is there any api to check from SPN ?

cgorringe:
Would if we did `return (hostname === 'archive.org') || (hostname === 'web.archive.org')`? I think that should allow for `mastodon.`, `blog.`, and `help.` to be archivable. (I haven't tested this.)

Alternately, could simply `return false` to accept all archive domains to test it rather than remove all the calls to `isArchiveUrl()`.

These are just suggestions. I do think there is some benefit in having the UI display "URL not supported" in known cases. I'm not sure what is the best solution.

cgorringe:
I'd recommend keeping it for `web.archive.org`.
When I tested it set to return false, the Wayback Count badge had some issues while viewing archived pages.

vbanos:
@anishsarangi please check again the SPN public API error codes.
https://docs.google.com/document/d/1Nsv52MvSjbLb2PCpHlat0gkzw0EvtSgpKHu4mk0MnrA/edit#heading=h.2bu3y5mtzecu
If a URL is blocked the API will instantly return a specific error code.
```
{"message":"This URL is in the Save Page Now service block list and cannot be captured. Please email us at \"info@archive.org\" if you would like to discuss this more.","status":"error","status_ext":"error:blocked-url"}
```

@cgorringe we may need to capture archive.org static pages, e.g. https://archive.org/about
You can't just block `archive.org`.

Again, the logic to allow/block a URL is too complicated. The SPN2 back-end is taking care of that.
You shouldn't do any blocking in the extension besides some basic URL validation.

anishsarangi:
@vbanos I think what @cgorringe suggests makes a valid point. I have just updated the code. Please have a look at it so that we can merge the PR.

vbanos:
OK, I've checked it and I think that its OK.

anishsarangi:
Okay i am merging this PR.

Issue #983, opened 2023-01-18
Clicking on the "Save Page Now" button does nothing at all


AlexanderRitter02:
**Describe the bug**
Clicking on the red "Save Page Now" button does nothing at all, no feedback, the button just stays looking the same.
This started happening after upgrading the extension to 3.2 and upgrading Firefox to 109.0, did these updates at the same time.

Strangely, on a fresh Firefox profile the isssue does not appear.
Can I please get help debugging this issue? E.g. can I somehow see **log output** from the extension or something similar?
Since this only seems to happen with my profile the scope of this issue may be minimal, but I would still appreciate _help or suggestions_ on how I could fix it. Thanks.

What I already tried:
- Reinstalling the extension
- Restarting Firefox and PC
- Clearing saved cookies and website data

**Environment:**
 - 3.2
 - Firefox 109.0
 - Windows 10 22H2

AlexanderRitter02:
By the way, when reinstalling the extension and clearing cookies, I was not prompted to login again.
This indicates that I'm not clearing all possible data related to the extension, how would I remove it completely?

cgorringe:
@AlexanderRitter02 Thanks for reporting this. Normally "Save Page Now" does work, but it may be at that instance our servers were not responding, which is why it would work when you tried it again.

To see the console log to help with debugging:
- Go here: Settings > Extensions & Themes > Upper-right Gear icon > Debug Add-ons
  - Or enter this URL: `about:debugging#/runtime/this-firefox`
- Select Extensions.
- Click on Inspect button right of Wayback Machine.
- Select "Console" tab.

The extension has its own persistent storage, which doesn't get cleared normally when clearing cookie data.
- To fully clear login cookies: Select the Logout button (lower-right corner of extension)
- To clear the cache, turn on Private Mode from the extension's settings.

That will clear out all of the extension's data.

AlexanderRitter02:
I don't believe it's the servers not responding, since it worked in another Firefox profile at the same time, and the issue still persists.

About logging out to delete data, I don't see a logout button:
![grafik](https://user-images.githubusercontent.com/42546705/214381675-374f28f4-21fe-4d30-a25a-48476fb46c0e.png)


Not sure what to do right now, since I can't wipe my entire Firefox profile, since I use the history for back reference and research.

AlexanderRitter02:
### Update: Problem solved ✔️ 
As stressed before, this was an issue with my profile.

[...]

PR #984, opened 2023-01-25
Cookie settings check

cgorringe:
Performs some checks on whether extension can access cookies and informs the user when they login to _Please Enable Cookies_.

- Catches case when `privacy.firstparty.isolate` variable in Firefox is set true. See #983
- Catches when All Cookies are blocked in browser settings. (which usually also prevents websites from working correctly.)

### DIFF
--- a/safari/Wayback Machine.xcodeproj/project.pbxproj
+++ b/safari/Wayback Machine.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 50;
+	objectVersion = 54;
 	objects = {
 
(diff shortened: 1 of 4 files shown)


saiqulhaq:
I am having this issue with container extension enabled

cgorringe:
I don't know what "container extension" enabled means.

In Firefox Privacy Settings: Strict Mode shouldn't cause issues normally. Logging in should only not work under Custom and ALL Cookies blocked. Other cookie settings should work. This PR will fix the issue by providing a warning to turn on cookies.

cgorringe:
Tested in latest Chrome, Firefox, Edge, and Safari.

PR #991, opened 2023-04-13
Update dependancies and package.json file

anishsarangi:
1. Updated the package that needs updating
2. Move `bootstrap` and `fontawesome-free` from devDependencies to dependencies since they are used at runtime.
3. Move `url-loader` from dependencies to devDependencies.
4. Fix an warning issue for `bootstrap.css.map System error: net::ERR_FILE_NOT_FOUND `

### DIFF
--- a/.github/workflows/continuousIntegration.yml
+++ b/.github/workflows/continuousIntegration.yml
@@ -7,6 +7,6 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2
       with:
-        node-version: 12
+        node-version: 16
     - run: npm install
     - run: npm test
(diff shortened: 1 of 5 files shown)


cgorringe:
We're not using webpack anymore to create a build.js script for production. The javascript libraries used are manually placed in `webextension/libs` and the html points to those files (which are not updated when running`npm run-script build`) and the reason for this is to make sure the libraries are readable and non-minified to pass code review. Now this can be fixed to work with webpack with some extra effort (and being careful about possible breaking changes in newer versions of the libraries) but the easy route is to probably just move everything to `devDependencies` and empty `dependencies`. (I have not test this)

anishsarangi:
We still need `jquery` as a dependancy for the testing. So I am moving out all the other dependancy except `jquery` from dependencies.

PR #996, opened 2023-08-16
Fix context yellow button to show correctly

anishsarangi:
Instead of opening the context url directly, this code shows the context button correctly and it redirects to the archived version of the url where the yellow banner exists.


anishsarangi:
Fixed in #998 . So closing this

PR #1023, opened 2024-04-02
fix: Google Play App link fixed.

ramazansancar:


anishsarangi:
Thank you for fixing it. Merging.

PR #1025, opened 2024-04-12
Manifest v3 upgrade

anishsarangi:
Upgraded the code to support manifest version 3


anishsarangi:
Fixed in #1042.
Closing it.

PR #1032, opened 2024-06-05
Explicitly specify which version of node to use

veyndan:
While going through the section [How to set up the development environment?](https://github.com/internetarchive/wayback-machine-webextension/blob/master/CONTRIBUTING.md#how-to-set-up-the-development-environment), I found I was unsure which version of node I should be using. By convention, this is specified in a file name `.node-version`, which almost all Node.js version managers [automatically interpret](https://github.com/shadowspawn/node-version-usage).

I'd updated the CI to point to this file as well, such that the CI and local development experience is more closely aligned.


cgorringe:
Thanks for pointing this out

PR #1034, opened 2024-06-05
Properly capitalize "Chrome"

veyndan:


cgorringe:
Thanks, I'll wait a bit in case you find any more words to update...

veyndan:
That was the only incorrect usage of "chrome" that I could find. I've finished reading the contributing documentation and I'm having a look at the code now, so probably won't be submitting any new word corrections (for a little while at least)!

PR #1066, opened 2025-03-31
Position new tabs next to the current tab

rdonaldwilson:
This commit makes it so that if features are set to open in new tabs then the tab is positioned next to the current tab instead of the current behaviour which always positions the tab at the end.


vbanos:
I'm sorry but we'll go with the default browser behavior

PR #1084, opened 2025-10-31
Upgrade stylelint

veyndan:
We're behind a few major versions of stylelint and stylelint-config-standard, so just bumping them.


vbanos:
Thank you but I just did a general dependency upgrade including this one
https://github.com/internetarchive/wayback-machine-webextension/pull/1087

PR #1089, opened 2025-11-02
Display extra SPN system status when capture fails

vbanos:
Check SPN system to detect potential general issues (e.g. SPN is overloaded) or user account issues (e.g. the user did too many requests today) and display extra info.


vbanos:
image
Example msg display.

PR #1093, opened 2025-11-19
Fix: Use Standard HTML Disabled Attribute and Bootstrap 5 Compatibility

anishsarangi:


vbanos:
@anishsarangi the attribute `disabled` is a standard HTML attribute.
What does it have to do with bootstrap compatibility?

If you want to style disabled buttons, you can use the `:disabled` CSS selector to do it.
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:disabled

PR #1100, opened 2026-01-11
Handle two exceptions, update JavaScript methods

Big-Blue-Gnu:
This version of the extension introduces improvements to make it work better on mobile devices, as my debugging efforts show. It adds two try-catch blocks to account for browsers that do not support certain APIs, such as context menus and windows. Notably, this fixes the error of the welcome page not closing on accepting the terms, and it should also hide the extension and display a list of excluded URLs in a separate tab.

More generally, the changes also replace deprecated JavaScript methods with recommended equivalents. At least on Firefox for Android, there have been no regressions to note as a result of the above improvements; they merely fix a fraction of the problems preventing the extension's wide availability on mobile browsers.


CookieSnippets:
The whole stuff with stripArchivePrefix() should be removed and left as it was. There is an obvious reason why archived pages are not allowed to be Saved again.
Also, if the intent was the possibility to re-save pages you come across on archive.org, all this stripping is completely unneeded if you simply strip the prefix ONCE in savePageNow().
👎👎

Big-Blue-Gnu:
I realize I should have done what I did in a separate branch. I did not realize that changes to the current branch would affect the pull request. I will drop the commits in the current branch and reapply them separately.

Also, it's nice to know that I took the hard way with `stripArchivePrefix()`, not realizing that all this could have been done simply inside `savePageNow()`. That was indeed the intent of it all: to allow users viewing archived versions of URLs to save the original while still inside the viewer. I'm sorry for the way I handled it. It was clumsy, and I should have just branched it to begin with and not relied on the master branch.

Big-Blue-Gnu:
Now I really botched it. I dropped all the new commits and attempted to undo all the subsequent commits that I merged into my fork, just to keep this pull request simple, but now I have 25 extra commits that I cannot seem to remove. I'm not sure what gives. At least I have the strip prefix patch [separately](https://github.com/Big-Blue-Gnu/wayback-machine-webextension/tree/patch-1).

Big-Blue-Gnu:
The guys at WineHQ were not kidding when they told me to always commit to separate branches. Here it seems, each time I try to revert to what it was before I worked on a new patch, I end up making it messier. Worse, I am not sure what I am doing wrong. I see all those merge conflicts when attempting to drop all irrelevant commits, but no indication of what to do about it. At this rate, until someone clarifies what I need to do to clean up this pull request, I give up.

Big-Blue-Gnu:
Seriously, what am I doing wrong? I clone my fork, then enter the editor with `rebase -i`, and when I attempt to drop the irrelevant commits, I end up needing to resolve merge conflicts. First of all, what is with the conflicts? Secondly, how is it that, each time, I skip straight to `git push -f`, I end up only adding more commits into my pull request? I am really confused at this point, and I would like to speak with the developers about how to revert to what it was before implementing my other patch (which is now viewable [here](https://github.com/Big-Blue-Gnu/wayback-machine-webextension/tree/patch-1)).

cgorringe:
Well, it's a good idea to create a new branch from master before making changes. This may be too much to look through for me to review as-is.

(Sorry I haven't reviewed code here in a while. I'm not currently employed at the Archive but have reviewed code here in the past and may get to doing it eventually.)

PR #1107, opened 2026-03-11
Fix Bootstrap 5 compatibility: disabled buttons, progress bar, and SPN  logged-out state

anishsarangi:
Fix Bootstrap 5 compatibility: disabled buttons, progress bar, and SPN  logged-out state
  - Preserve custom button colors in disabled state by adding :disabled base
    selectors to override Bootstrap 5's .btn:disabled which uses undefined CSS
    variables that resolve to transparent, making buttons invisible in light
  mode
  - Set .progress background to transparent so .btn-red parent shows through
    instead of Bootstrap 5's light gray secondary-bg
  - Restore progress bar stripe animation by adding explicit height, larger
    background-size, and original keyframes that Bootstrap 5 changed
  - Fix Save Page Now not working when logged out: remove auth_check gate in
    savePageNowChecked so savePageNow is always called with the correct flag

### DIFF
--- a/webextension/css/common.css
+++ b/webextension/css/common.css
@@ -102,22 +102,28 @@ button:disabled {
     box-shadow: none !important;
 }
 
-/* Preserve original button colors for disabled state - override hover/active states */
+/* Preserve original button colors for disabled state - override Bootstrap 5's
+   .btn:disabled which uses undefined CSS vars that resolve to transparent */
+.btn-red:disabled,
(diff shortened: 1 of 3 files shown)


mitar:
Thanks!

When will this be released and available for Firefox?

anishsarangi:
We are working on it. We will have a new release for Firefox very soon.

Issue #1114, opened 2026-03-28
404 "pop-up in page" banner not showing


Cicici-Shi:
## Bug Description

The "pop-up in page" option under "404 Not Found, etc..." in Settings does not work. The toolbar red dot appears correctly, but the in-page banner never shows.

## Steps to Reproduce

1. Enable "404 Not Found, etc..." in Settings
2. Check "pop-up in page"
3. Visit https://www.whitehouse.gov/energy/climate-change (returns 404)

**Expected:** An in-page popup with "View Archived Version" link.
**Actual:** No popup appears.

## Environment

- Extension version: 3.4.8
- Browser: Chrome 146.0.7680.165 (arm64)

PR #1115, opened 2026-03-28
Fix SyntaxError in archive.js preventing 404 popup banner

Cicici-Shi:
Fix for issue #1114. Optional chaining (`?.`) on the left-hand side of an assignment is [invalid JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#invalid_optional_chaining), causing `archive.js` to fail to parse and the "pop-up in page" banner to never appear.


cgorringe:
Thanks.

PR #1135, opened 2026-09-04
Open new tabs next to the current one

### LINKED ISSUE #1134: Open Archived Versions/Calenders Next to the Current Tab
**Is your feature request related to a problem/ issue? Please describe.**
Archived pages and calenders, whether from the context menu on page or the extension popup, opens as a new last tab in the window. This disrupts how new tabs usually open next to the current one, keeping related tabs together. Furthermore, closing the tab switches to the unrelated next last tab. This is most frustrating on windows with a lot of unrelated tabs.

Illustration (***emphasis***: current tab):
|  |  |  |  | Action |
|---|---|---|---|---|
| ***Tab 1*** | Tab 2 | Tab 3 |  |  |
| Tab 1 | Tab 2 | Tab 3 | ***Tab 1 (arc)*** | Open an archived version of Tab 1 |
| Tab 1 | Tab 2 | ***Tab 3*** |  | Close the archived version |

**Describe the solution you'd like**
Change it so the extension opens tabs next to the current one.

|  |  |  |  | Action |
|---|---|---|---|---|
| ***Tab 1*** | Tab 2 | Tab 3 |  |  |
| Tab 1 | ***Tab 1 (arc)*** | Tab 2 | Tab 3 | Open an archived version of Tab 1 |
| ***Tab 1*** | Tab 2 | Tab 3 |  | Close the archived version |

**Describe alternatives you've considered**
N/A

Naruyoko:
Fixes #1134

The initial implementation changes the behavior of _all_ new tabs. I think this is an acceptable change, but if the previous behavior is desired in some situations, an optional parameter/configuration to the utility function may be added.


webextension/scripts/utils.js
```diff
@@ -650,8 +650,10 @@ function openByWindowSetting(url, op = null, cb) {
 
 function opener(url, option, callback) {
   if (option === 'tab' || option === undefined) {
-    chrome.tabs.create({ url }, (tab) => {
-      if (callback) { callback(tab.id) }
+    chrome.tabs.query({ active: true, currentWindow: true }, (tabs) => {
+      chrome.tabs.create({ url, index: tabs?.[0].index + 1 ?? 0 }, (tab) => {
```
CookieSnippets: index: tabs?.[0].index + 1 ?? 0

**This does not do what you think it does.**
Because you do +1 before ??,  the fallback 0 never triggers.

For example, if tabs?.[0].index is undefined or whatever, you get :
undefined + 1 ?? 0, which is NaN, en thus NaN ?? 0 gives again Nan not 0.

The correct and safe index calculation is:
**index: (tabs?.[0]?.index ?? -1) + 1**

This gives 0 when the first tab's index is null/undefined (or the tab itself is missing).

